PR #1, opened 2019-08-27
Rename to `ocm`

jhernand:
This patch renames the tool to `ocm` and adjusts the code for the new
home at `github.com/openshift-online/ocm-cli`.


jhernand:
FYI: @ALimaRedHat

PR #3, opened 2019-08-27
Add `build_deploy.sh` script

jhernand:
This patch add a new `build_deploy.sh` script that builds the binaries
of the project for all the supported platforms (currently _Linux_ and
_macOS_).


jhernand:
[test]

PR #8, opened 2019-09-07
Add ability to open the console URL in browser

cblecker:


cmd/ocm/cluster/login/login.go
```diff
@@ -47,13 +49,8 @@ var Cmd = &cobra.Command{
...
 func init() {
 	flags := Cmd.Flags()
-	flags.StringVarP(
-		&args.user,
-		"username",
-		"u",
-		"",
-		"Username, will prompt if not provided",
-	)
+	flags.StringVarP(&args.user, "username", "u", "", "Username, will prompt if not provided")
```
jhernand: In other parts of the code when we need to specify an user we use just `user`. For example in the top level command (although that option is deprecated). Can we use `user` instead of `username` here as well?

I would also prefer if we stick to the convention of putting each line of the cals to `flags.WhaveverVar(...)` in a different line. It makes it easier to write longer help texts.

cmd/ocm/cluster/login/login.go
```diff
@@ -47,13 +49,8 @@ var Cmd = &cobra.Command{
...
+	flags.BoolVarP(&args.console, "console", "", false, "Open the OpenShift console for the cluster")
```
jhernand: I think this help text should explicitly indicate that this will open the console in a browser.

cmd/ocm/cluster/login/login.go
```diff
@@ -106,30 +103,41 @@ func run(cmd *cobra.Command, argv []string) error {
 			total, argv[0], len(clusters),
 		)
 	}
-	var clusterid, clusterName, url string
+	var clusterid, clusterName, apiURL, consoleURL string
```
jhernand: I'd suggest to have here a `cluster` variable:

```go
var cluster v1.Cluster
if len(clusters) == 1 {
        cluster = clusters[0]
        fmt.Printf(...)
} else {
        cluster, err = doSurvey(clusters)
        ...
}
```

cmd/ocm/cluster/login/login.go
```diff
@@ -106,30 +103,41 @@ func run(cmd *cobra.Command, argv []string) error {
...
+		apiURL = cluster.API().URL()
+		consoleURL = cluster.Console().URL()
 		clusterid = cluster.ID()
 		clusterName = cluster.Name()
 	}
 
-	if len(url) == 0 {
+	if args.console {
+		if len(consoleURL) == 0 {
+			return fmt.Errorf("Cannot find the console url for cluster: %s", clusterName)
```
jhernand: URL is an acronym so it should be upper case: `... console URL for ...`

cblecker:
@jhernand Feedback addressed! Thanks!

PR #17, opened 2019-09-26
Release 0.1.25

jhernand:
The more relevant changes in this version are the following:

- Added new `cluster create` command.

- Added support for `production`, `staging` and `integration` as values
  of the `--url` parameter.


jhernand:
FYI: @cblecker

PR #19, opened 2019-09-27
Add support for an enabled label on ClusterImageSets

jewzaam:
Label 'api.openshift.com/enabled' is optional.
If label is not found, version defaults to enabled=true.
If set and value is `false` the version is enabled=false.

https://jira.coreos.com/browse/SREP-2175


jewzaam:
wrong repo, ignore..

PR #20, opened 2019-09-30
Retrieve all enabled versions

jhernand:
Currently the `cluster create` command retrieves only the first page of
enabled versions. This works fine in environments with less than 100
versions, but in the integration environment there are already 116
versions. This results in rejecting the command when the user tries to
use any of the newer versions that aren't in the first page. To avoid
that this patch changes the command so that it will always retrieve all
the versions, even if there are more than 100.

### DIFF
--- a/cmd/ocm/cluster/create/create.go
+++ b/cmd/ocm/cluster/create/create.go
@@ -102,19 +102,16 @@ func run(cmd *cobra.Command, argv []string) error {
 	// Retrieve valid/default versions
 	versionList := sets.NewString()
 	var defaultVersion string
-
-	versionsResponse, err := cmv1Client.Versions().List().Search("enabled = 'true'").Send()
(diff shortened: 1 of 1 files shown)


jhernand:
@cblecker please review.

cmd/ocm/cluster/create/create.go
```diff
@@ -164,3 +161,26 @@ func run(cmd *cobra.Command, argv []string) error {
...
+	size := 100
+	for {
+		var response *cmv1.VersionsListResponse
+		response, err = collection.List().
+			Search("enabled = 'true'").
+			Page(page).
+			Size(size).
+			Send()
+		if err != nil {
+			return
```
cblecker: stylistically, would it be better to not use naked returns?

jhernand: Well, I agree it is a matter of style/taste. I like them.

PR #29, opened 2019-11-18
Allow bare `ocm login` to suggest the token page without extra noise

cben:
@jhernand please review.
This guides users towards using --token from https://cloud.redhat.com/openshift/token.
That page is not yet advertised in UI but I think we should, as that's the sane ways for customers to access the API?

```sh
$ ocm login
In order to log in it is mandatory to use '--token', '--user' and '--password', or '--client-id' and '--client-secret'.
You can obtain a token at: https://cloud.redhat.com/openshift/token .
See 'ocm login --help' for full help.
$ echo $?
1
```

Works with some flags too, as long as you didn't provide crendentials:

```sh
$ ocm login --url=staging
In order to log in it is mandatory to use '--token', '--user' and '--password', or '--client-id' and '--client-secret'.
You can obtain a token at: https://cloud.redhat.com/openshift/token .
See 'ocm login --help' for full help.
$ echo $?
1
```

(In my testing, the same token is valid for logging into production, staging, and integration.)

Full help:

```sh
$ ocm login --help
Log in, saving the credentials to the configuration file.
The recommend way is using '--token', which you can obtain at: https://cloud.redhat.com/openshift/token

Usage:
  ocm login [flags]

Flags:
      --.....
```

(and exactly same for `ocm help login`)

- I'd like to later improve the token page similarly to openshift console's — link it from (?) icon -> Command Line Tools, and reword it to advertise this CLI and suggest `ocm login --token`, only later discussing direct low-level usage.

https://jira.coreos.com/browse/SDA-1609

### DIFF
--- a/cmd/ocm/login/cmd.go
+++ b/cmd/ocm/login/cmd.go
@@ -35,6 +35,9 @@ var urlAliases = map[string]string{
 	"integration": "https://api-integration.6943.hive-integration.openshiftapps.com",
 }
 
+// #nosec G101
+const uiTokenPage = "https://cloud.redhat.com/openshift/token"
+
 var args struct {
 	tokenURL     string
 	clientID     string
@@ -51,8 +54,9 @@ var args struct {
 var Cmd = &cobra.Command{
 	Use:   "login",
 	Short: "Log in",
-	Long:  "Log in, saving the credentials to the configuration file.",
-	RunE:  run,
+	Long: "Log in, saving the credentials to the configuration file.\n" +
+		"The recommend way is using '--token', which you can obtain at: " + uiTokenPage,
+	RunE: run,
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


cmd/ocm/login/cmd.go
```diff
@@ -144,8 +147,16 @@ func run(cmd *cobra.Command, argv []string) error {
...
 	haveToken := args.token != ""
 	if !havePassword && !haveSecret && !haveToken {
-		return fmt.Errorf("In order to log in it is mandatory to use '--token', '--user' and " +
-			"'--password', or '--client-id' and '--client-secret'.")
+		// Allow bare `ocm login` to suggest the token page without extra noise.
+		fmt.Fprintf(
+			os.Stderr,
+			"In order to log in it is mandatory to use '--token', '--user' and "+
+				"'--password', or '--client-id' and '--client-secret'.\n"+
+				"You can obtain a token at: %s .\n"+
```
cben: Space before period is awkward but deliberate.
Not all terminal allow direct clicking of URLs, not all users know how to use that (e.g. Ctrl+click in gnome-terminal, or right click menu). The space makes it easier to grab the URL with regular mouse selection or double-click word selection.

Could trop the period altogether, but with the 2 other sentences ending in a period felt that would look awkward too.

cmd/ocm/login/cmd.go
```diff
@@ -144,8 +147,16 @@ func run(cmd *cobra.Command, argv []string) error {
...
+				"See 'ocm login --help' for full help.\n",
+			uiTokenPage,
+		)
+		return nil
```
jhernand: This should `os.Exit(1)` otherwise scripts that use the tool will assume it succeeded in this case when it actually failed.

cben: Thanks, done.

PR #30, opened 2019-11-18
Release 0.1.29

jhernand:
The more relevant changes in the new version are the following:

- Allow bare `ocm login` to suggest the token page without extra noise.


jhernand:
FYI: @cben

PR #37, opened 2019-12-12
Add RoleBindings and ResourceQuota shortcuts

rawsyntax:

PR #38, opened 2019-12-12
Add roles and skus shortcut

rawsyntax:


rawsyntax:
@jhernand a few more shortcuts I missed in #37, thanks for reviewing!

PR #40, opened 2019-12-16
adding new options for "ocm account users" command

rhdedgar:
Adding options needed by SRE Security for quarterly compliance review processes.

Now ocm gives us the ability to sift through the entirety of the user base and find those with specific roles in under a minute.

Every 3 months, we'll run a beefy query like the following:
ocm account users --roles="role1,role2,role3" --workers=100 --pagesize=100000

### DIFF
--- a/cmd/ocm/account/users/cmd.go
+++ b/cmd/ocm/account/users/cmd.go
@@ -25,29 +25,26 @@ import (
 
 	acc_util "github.com/openshift-online/ocm-cli/pkg/account"
 	"github.com/openshift-online/ocm-cli/pkg/config"
-	sdk "github.com/openshift-online/ocm-sdk-go"
 	amv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 )
 
-var (
-	pageIndex = 1
-	namePad   = 40
-)
-
 var args struct {
-	debug    bool
-	org      string
-	roles    []string
-	workers  int
-	pageSize int
+	debug bool
+	org   string
+	roles []string
 }
 
-// Cmd defines a new cobra Command
+// Cmd configures a new Cobra Command
 var Cmd = &cobra.Command{
 	Use:   "users",
 	Short: "Retrieve users and their roles",
 	Long:  "Retrieve information of all users/roles in the same organization",
 	RunE:  run,
+}
+
+type userModel struct {
+	userName string
+	userID   string
 }
 
 func init() {
@@ -70,18 +67,6 @@ func init() {
 		"roles",
 		[]string{},
 		"Role identifiers. Returns users with one or more of the specified roles. Multiple roles can be specified like: --roles=\"role1,role2,role2\".",
-	)
-	flags.IntVar(
-		&args.workers,
-		"workers",
-		1,
-		"Used with --roles. Number of workers to which we distribute the load. Queries run faster but use more CPU.",
-	)
-	flags.IntVar(
-		&args.pageSize,
-		"pagesize",
-		100,
-		"Size of page to return from the server. Larger page sizes equal faster search times with --roles.",
 	)
 }
 
@@ -112,10 +97,16 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 	defer connection.Close()
 
+	// needed variables:
+	pageSize := 100
+	pageIndex := 1
+	namePad := 40
 	searchQuery := ""
 
-	// Organization to search in case one was not provided.
-	// Only used when we're not already searching all orgs.
+	if args.org != "" {
+		searchQuery = fmt.Sprintf("organization_id='%s'", args.org)
+	}
+	// Organization to search in case one was not provided:
 	if args.org == "" && len(args.roles) == 0 {
 		// Get organization of current user:
 		userConn, err := connection.AccountsMgmt().V1().CurrentAccount().Get().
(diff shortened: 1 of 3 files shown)


rhdedgar:
Adding active maintainer @jhernand to the PR.

Please let me know if any changes are needed prior to merge. 

Thank you!

jhernand:
I am a concerned about the approach taken here. Looks to me that there is performance problem in the original implementation of the `account users` command and that you are using the large number of workers and the large page size to work around that performance problem and get the result in a reasonable time.

Both the number of workers and the large page size are potential abuses of the server. With 100 workers you will be probably using all the database connections available in the server side, leaving almost no room for other simultaneous requests.

With 100000 as the page size you are forcing the server to load those 100000 records in memory before starting to send them to the client, which will consume quite an amount of memory.

I'd prefer to look at the original problem and try to solve it. One of the potential performance problems in the initial implementation of the `account users` implementation is that it sends a separate request to retrieve the role bindings for each account using the search query `account_id = 123`. A good way to improve that is to use the `in` operator in the search query to retrieve the role bindings for multiple accounts simultaneously. For example, you would re-write the `GetAccountForUser` like this:

```go
func GetRolesFromUsers(accounts []*amv1.Account, conn *sdk.Connection) (result map[*amv1.Account][]*amv1.Role, error error) {
        // Prepare the result:
        result = map[*amv1.Account][]*amv1.Role{}

        // Prepare a map of accounts indexed by identifier:
        accountsMap := map[string]*amv1.Account{}
        for _, account := range accounts {
                accountsMap[account.ID()] = account
        }

        // Prepare a map of roles indexed by identifier:
        rolesMap := map[string]*amv1.Role{}

        // Prepare a query to retrieve all the role bindings that correspond to any of the
        // accounts:
        ids := &bytes.Buffer{}
        for i, account := range accounts {
                if i > 0 {
                        ids.WriteString(", ")
                }
                fmt.Fprintf(ids, "'%s'", account.ID())
        }
        query := fmt.Sprintf("account_id in (%s)", ids)

        // Prepare the request:
        request := conn.AccountsMgmt().V1().RoleBindings().List().Search(query)
        index := 1
        size := 100
        for {
                // Send the request:
                response, err := request.Page(index).Send()
                if err != nil {
                        err = fmt.Errorf("can't retrieve roles: %v", err)
                        return
                }

                // Loop through the results and save them:
                response.Items().Each(func(item *amv1.RoleBinding) bool {
                        account := accountsMap[item.Account().ID()]
                        role := rolesMap[item.Role().ID()]
                        if role == nil {
                                role = item.Role()
                                rolesMap[role.ID()] = role
                        }
                        roles := result[account]
                        if !roleInList(roles, role) {
                                roles = append(roles, role)
                                result[account] = roles
                        }
                        return true
                })

                // Break the loop if the page size is smaller than requested, as that indicates
                // that this is the last page:
                if response.Size() < size {
                        break
                }
                index++
        }

        return
}
```

Then you could call that method with a block of accounts instead of just with one account. I think that could give you a performance improvement enough to avoid the multiple workers and the huge page size. Can you explore that?

rhdedgar:
Can do, I'll look into a workaround. 

One of the problems has been developing this around the "black-box" API, do you have any documents that you could share? Or is the expectation for developers to grok api.openshift.com?

From random queries I've run, there aren't even 100,000 accounts in total, more like 30,000 (that command was just an example). But things like database connection limits, etc. are opaque to me.

https://github.com/openshift-online/ocm-sdk-go is also pretty light on documentation.

jhernand:
The reference documentation of the API is in https://api.openshift.com. For example the description of the `search` parameter of the `role_bindings` collection is [here](https://api.openshift.com/?urls.primaryName=Accounts%20management%20service#/default/get_api_accounts_mgmt_v1_role_bindings).

That can indeed be improved. What specific information would you like to see there to make your work easier? Can you open a JIRA issue describing that?

The reference documentation for the SDK is [here](https://godoc.org/github.com/openshift-online/ocm-sdk-go) and there is a collection of examples [here](https://github.com/openshift-online/ocm-sdk-go/tree/master/examples).

Your feedback is important, so again I'd appreciate if you can open GitHub issues asking for specific improvements to the documentation of the SDK.

We don't have any other documentation.

rhdedgar:
Ok thanks for the suggestions! I've revised the initial commit with the new bulk role retrieval workflow. The command runs at an acceptable speed except when it hits one particular org.

There seems to be more than 243,600 "openshift-ci-test" users in the "1Hjy0dhTVwN0oQ29KY6J47w0K8n" organization that **really** slow down the speed of any queries that include it.

If you try this with the current release version of ocm, it takes a very long time to return:
ocm account users --org=1Hjy0dhTVwN0oQ29KY6J47w0K8n

Adding Clayton Coleman (the owner of that org) to the PR to check what that's about.
@smarterclayton are all these accounts needed, and can they be cleaned up?

rhdedgar:
@jhernand revisiting this now that everyone is back from PTO. Could you please take a look when you get the chance?

@smarterclayton did you ever get time to assess if any of those duplicate accounts in the openshift-ci-test org can be cleaned up?


Thanks

cmd/ocm/account/users/cmd.go
```diff
@@ -55,6 +62,12 @@ func init() {
 		"", // Default value gets assigned later as connection is needed.
 		"Organization identifier. Defaults to the organization of the current user.",
 	)
+	flags.StringSliceVar(
+		&args.roles,
+		"roles",
+		[]string{},
+		"Role identifiers. Returns users with one or more of the specified roles. Multiple roles can be specified like: --roles=\"role1,role2,role2\".",
```
jhernand: @rhdedgar The tests are failing because this line is very long. Change this so that the line is at most 100 characters and I will then merge.

rhdedgar: Ok, this is now split into two lines (each < 100 characters). Thank you

PR #41, opened 2020-01-01
[SDB-840] Add whatami and whatcani commands

gshilin-sdb:


cmd/ocm/whatami/cmd.go
```diff
@@ -0,0 +1,88 @@
...
+	names := make([]string, 0, listResponse.Size())
+	listResponse.Items().Each(func(role *amsv1.Role) bool {
+		name, _ := role.GetName()
+		names = append(names, fmt.Sprintf("\"%s\"", name))
+
+		return true
+	})
+	name := strings.Join(names, ",\n    ")
+	fmt.Printf("    %s\n", name)
+	fmt.Printf("  ]\n}\n")
```
jhernand: If you are going to generate a JSON document that looks like the original one returned by the server it may be better to just dump the original JSON returned by the server. Is there any reason to not do that?

tiwillia:
@abhgupta can we get your input on whether we _want_ these features in the ocm cli? Do we want to expose the roles and permissions a user has to them?

abhgupta:
I have reached out to Jake and Cameron to get their inputs on SDB-299 and SDB-840

igoihman:
closing, feel free to re-open in the future if the change will be needed

PR #42, opened 2020-01-02
Add commands `ocm whatami` and `ocm whatcani`

gshilin-sdb:


jhernand:
This will be replaced by a different release.

Issue #48, opened 2020-01-14
Feature request – global flag for overriding default config file path


mmazur:
Just like `--kubeconfig` in `kubectl`.

mmazur:
Env var is enough.

PR #50, opened 2020-01-16
Add OCM_CONFIG env var support

mmazur:
Implements #48 with a global var. I'd also do `--config`, but there's no existing global param support in the code, so figuring out the proper structures is non–trivial for someone new to the code.

This, on the other hand, is a very small patch.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,7 +1,7 @@
 = OCM API Command Line Tools
 
 This project contains the `ocm` command line tool that simplifies the use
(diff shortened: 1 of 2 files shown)


mmazur:
Just wanted to get an initial ack first. If the patch looks ok, I'll fix the ci tests and add the docs.

mmazur:
@jhernand docs added, linter errors fixed, ready for final inspection.

PR #53, opened 2020-02-03
Release 0.1.35

jhernand:
The more relevant changes in the new version are the following:

- Display quota so it supports add-ons.


jhernand:
FYI: @aweiteka

PR #58, opened 2020-02-19
Add MacOS installer command

wgordon17:


README.adoc
```diff
@@ -17,7 +18,15 @@ Copr]. To enable that repository and install the tool use the following commands
 This will install the `ocm` command and will keep it updated using the same
 mechanism used to update all the other packages of the distribution.
 
-If you are not using one of this distributions or you don't want to use the RPM
+=== MacOS Brew
+
+....
+# brew install ocm
```
jhernand: In the other examples of commands in this file we used `$ ...` to suggest that this is a command to execute as a regular user. When I see `# ...` my first tought is that this needs to be executed as root instead. Is that true? I thought that that "brew" command can be executed as a regular user. Is that correct? If so can we change the example to `$ brew install ocm`?

wgordon17: Yeah, this can be `$`, sorry, I was just basing my `#` usage from the lines above ¯\_(ツ)_/¯

PR #65, opened 2020-03-17
Add support for expiration in ocm cluster create

cblecker:
Now that we have maximum expiration support in the OCM API, this makes it easy to specify an expiration date via the CLI for a cluster you are creating. You can specify an exact date in RFC3389 format, or you can choose a duration (like 8h or 2d).

### DIFF
--- a/cmd/ocm/cluster/create/create.go
+++ b/cmd/ocm/cluster/create/create.go
@@ -18,6 +18,7 @@ package create
 
 import (
 	"fmt"
+	"time"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
@@ -29,11 +30,13 @@ import (
 )
 
 var args struct {
-	parameter []string
-	header    []string
-	region    string
-	version   string
-	flavour   string
+	parameter         []string
+	header            []string
+	region            string
+	version           string
+	flavour           string
+	expirationTime    string
+	expirationSeconds time.Duration
 }
 
 // Cmd Constant:
@@ -67,10 +70,28 @@ func init() {
 		"osd-4",
 		"The OCM flavour to create the cluster with",
 	)
+	fs.StringVar(
+		&args.expirationTime,
+		"expiration-time",
+		args.expirationTime,
+		"Specified time when cluster should expire (RFC3339). Only one of expiration-time / expiration may be used.",
+	)
+	fs.DurationVar(
+		&args.expirationSeconds,
+		"expiration",
+		args.expirationSeconds,
+		"Expire cluster after a relative duration like 2h, 8h, 72h. Only one of expiration-time / expiration may be used.",
+	)
 }
 
 func run(cmd *cobra.Command, argv []string) error {
 	var err error
+	var expiration time.Time
+
+	// Validate options
+	if len(args.expirationTime) > 0 && args.expirationSeconds != 0 {
+		return fmt.Errorf("at most one of `expiration-time` or `expiration` may be specified")
+	}
 
 	// Load the configuration file:
 	cfg, err := config.Load()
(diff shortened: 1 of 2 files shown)


cblecker:
ping @jhernand

cmd/ocm/cluster/create/create.go
```diff
@@ -106,6 +128,21 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 	clusterName := argv[0]
 
+	// Parse the expiration options
+	if len(args.expirationTime) > 0 {
+		t, err := parseRFC3339(args.expirationTime, metav1.Now)
+		if err != nil {
+			return err
```
cben, line 135: For user friendliness, maybe wrap using Errorf to explain which flag had the problem.

cmd/ocm/cluster/create/create.go
```diff
@@ -164,8 +201,13 @@ func run(cmd *cobra.Command, argv []string) error {
 		Version(
 			cmv1.NewVersion().
 				ID(clusterVersion),
-		).
-		Build()
+		)
+	if args.expiration != nil {
+		clusterBuild = clusterBuild.ExpirationTimestamp(
+			args.expiration.Time,
```
cben, line 207: Do you need the metav1.Time wrapping if in the end  you extract .Time field?

cmd/ocm/cluster/create/create.go
```diff
@@ -67,11 +72,28 @@ func init() {
...
+		"expiration-time",
+		args.expirationTime,
+		"Specified time when cluster should expire (RFC3339). Only one of expiration-time / expiration may be used.",
+	)
+	fs.DurationVar(
+		&args.expirationSeconds,
+		"expiration",
+		args.expirationSeconds,
+		"Expire cluster after a relative duration like 2h, 8h, 3d. Only one of expiration-time / expiration may be used.",
+	)
```
vkareh, line 86: Is there any specific use-case for supporting both flags?

I'm wondering what people would use the most and settling on that - you can always do, for example
```
ocm cluster create --expiration-timestamp=$(date -Iseconds -d+7days)
```
If the goal of supporting both formats is for scripting purposes, then it can be done like that example. Otherwise we can just document one of them that we're more likely to use. I would personally use this a format lot more often:
```
ocm cluster create --expiration=3d
```
but the timestamp approach might be a lot more scriptable, and so might make more sense to support that one.

cmd/ocm/cluster/create/create.go
```diff
@@ -29,11 +31,14 @@ import (
...
-	version   string
-	flavour   string
+	parameter         []string
+	header            []string
+	region            string
+	version           string
+	flavour           string
+	expirationTime    string
+	expirationSeconds time.Duration
+	expiration        *metav1.Time
```
vkareh, line 41: I'm not sure `expiration` should be part of `args`, since it's only ever assigned and used inside `run()` we might as well define it locally.

cmd/ocm/cluster/create/create.go
vkareh, line 207: Something seems to be wrong here (perhaps a bug in the CS API, but it could be in the spf13/flag module):
```
$ ocm cluster create --expiration=7d mycluster
Error: invalid argument "7d" for "--expiration" flag: time: unknown unit d in duration 7d
```

cblecker: Tracked this down, and it's actually `time.Duration` not supporting days. I don't quite feel like writing a custom parser for it, so I've updated the help text to suggest using `72h` instead of `3d`.

cblecker: I was using it for nil checking, but I can use `IsZero()` instead.

cblecker:
I also added a commit that will expose the expiration timestamp (and change the creation timestamp format) so that it looks like:
```
Created:     2020-04-06T17:10:00Z
Expiration:  2020-04-07T03:00:00Z
```

cmd/ocm/cluster/create/create.go
vkareh: I see now, makes sense. Thanks for the explanation :+1:

cmd/ocm/cluster/create/create.go
```diff
@@ -106,6 +127,20 @@ func run(cmd *cobra.Command, argv []string) error {
...
+	// Parse the expiration options
+	if len(args.expirationTime) > 0 {
+		t, err := parseRFC3339(args.expirationTime)
+		if err != nil {
+			return fmt.Errorf("unable to parse expiration time: %s", err)
+		}
+
+		expiration = t
+	}
+	if args.expirationSeconds != 0 {
```
vkareh, line 139: Wondering if we could use `time.ParseDuration()`

cblecker: @vkareh ParseDuration takes a string, but the duration is already being parsed by https://pkg.go.dev/github.com/spf13/pflag?tab=doc#FlagSet.DurationVar, so at this point in the code it's a time.Duration already

vkareh: of course, thanks!

PR #66, opened 2020-03-19
Add support for specifying cloud provider

cblecker:
This adds support for specifying a cloud provider to `ocm cluster create` via a `--provider` flag

### DIFF
--- a/cmd/ocm/cluster/create/create.go
+++ b/cmd/ocm/cluster/create/create.go
@@ -35,6 +35,7 @@ var args struct {
 	region            string
 	version           string
 	flavour           string
+	provider          string
 	expirationTime    string
 	expirationSeconds time.Duration
 }
@@ -56,7 +57,7 @@ func init() {
 		&args.region,
 		"region",
 		"us-east-1",
-		"The AWS region to create the cluster in",
+		"The cloud provider region to create the cluster in",
 	)
 	fs.StringVar(
 		&args.version,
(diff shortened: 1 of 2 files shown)


cmd/ocm/cluster/create/create.go
```diff
@@ -93,6 +100,9 @@ func run(cmd *cobra.Command, argv []string) error {
 	if len(args.expirationTime) > 0 && args.expirationSeconds != 0 {
 		return fmt.Errorf("at most one of `expiration-time` or `expiration` may be specified")
 	}
+	if args.region == "us-east-1" && args.provider != "aws" {
+		return fmt.Errorf("if specifying a non-aws cloud provider, region must be set")
```
staebler: Probably not too big of a deal, but this message would be misleading if the user explicitly set the region to "us-east-1".

cblecker: That's true, but I wanted to maintain the current backward compatible aws defaults, *and* I want cobra to display them in help text. I'm open to suggestions on how to better word this, or check for this use case.

cblecker: I had another look at this.. it will only fire if the provider is NOT aws, and `us-east-1` isn't a valid region in gcp. I amended the language to add the word valid if someone actually specifies the combination of `--provider=gcp --region=us-east-1` manually.

cblecker:
/cc @vkareh @cben @staebler 
Would one of you mind having a look at this one too when you get a chance? Thank you!

vkareh:
LGTM

PR #69, opened 2020-04-03
ocm login --use-config

2uasimojo:
This allows you to log in using the values from your config without
overwriting it. This lets you use a refresh token indefinitely if
desired, reducing/eliminating "token expired" errors while doing other
stuff like ocm cluster login.


2uasimojo:
[test]

2uasimojo:
[test]

igoihman:
@vkareh PTAL

2uasimojo:
/hold
bleh, this doesn't work

cben:
This would be neat for internal use as we have several production/staging/integration envs to log into, all using same SSO!

For "token expired" use cases, I'm not sure we need it.  Most commands already write refreshed tokens back to config file, #153 discusses some corner cases to improve. 
@2uasimojo assuming we fix up all commands to save refreshed tokens, do you still see this as useful?

2uasimojo:
> This would be neat for internal use as we have several production/staging/integration envs to log into, all using same SSO!
> 
> For "token expired" use cases, I'm not sure we need it. Most commands already write refreshed tokens back to config file, #153 discusses some corner cases to improve.
> @2uasimojo assuming we fix up all commands to save refreshed tokens, do you still see this as useful?

To be honest, I'm not sure. So let's back up to my use case.

What was frustrating me was being able to run ocm commands for "a while" (whatever the short-lived token timeout is) but eventually failing and having to `ocm login` again. Especially since it _looked_ like there was a way for the config to store my offline token. That's what I really want: to be able to set up my config with my offline token, and then run arbitrary `ocm` commands forever, without ever having to log in explicitly again*. In pseudocode terms, I want every auth-needing invocation of ocm to do this:

```
if !config.ephemeral_token || expired(config.ephemeral_token):
    if config.offline_token:
        config.ephemeral_token = new_ephemeral_token(config.offline_token)
        write_config(config)  # Important: this preserves config.offline_token
    else:
        { other ways of authenticating, like user/pass CLI args }
```

\* Understanding that we have separate environments, I guess I'm okay needing to log in explicitly when I need to switch from one to the other. But even there, why should I need to, since the offline token is the same? Heck, what would be _really_ neat is if the config were able to maintain separate sections, each with its short-lived token, to minimize SSO calls even when I'm thrashing among envs.

Nosense363 (approved):

PR #72, opened 2020-04-14
Add cloud provider to default columns

cblecker:


cmd/ocm/cluster/list/list.go
```diff
@@ -72,10 +72,8 @@ func init() {
 	fs.StringVar(
 		&args.columns,
 		"columns",
-		"id, name, api.url, openshift_version, region.id, state",
-		"Specify which columns to display separated by commas, path is based on Cluster struct i.e. "+
-			"id,name,api.url,openshift_version,region.id,state"+
```
cblecker, line 77: Removed this as the default value already displays in the help:
```
$ ./ocm cluster list --help
List clusters by ID and Name

Usage:
  ocm cluster list [flags] [partial cluster ID or name]

Flags:
      --columns string          Specify which columns to display separated by commas, path is based on Cluster struct (default "id, name, api.url, openshift_version, cloud_provider.id, region.id, state")
```

cmd/ocm/cluster/list/list.go
```diff
@@ -131,7 +129,7 @@ func run(cmd *cobra.Command, argv []string) error {
 	colUpper := strings.ToUpper(args.columns)
 	colUpper = strings.Replace(colUpper, ".", " ", -1)
 	columnNames := strings.Split(colUpper, ",")
-	paddingByColumn := []int{34, 45, 70, 60, 15}
+	paddingByColumn := []int{34, 40, 60, 20}
```
cblecker, line 132: Fixed up and tightened the default padding for the default columns list

PR #76, opened 2020-05-17
Add shell completion for resources

mrbarge:
This PR adds shell auto-completion for resources which can be used by the "get", "delete", "patch" and "post" commands. This reduces the need for OCM users to memorize API URLs, and promotes the use of the already-existing resource shortcuts.

For example, once the bash completion script is generated and sourced:
`ocm get a<tab><tab>`
will output:
`account   accounts  acct      accts`

### DIFF
--- a/cmd/ocm/delete/cmd.go
+++ b/cmd/ocm/delete/cmd.go
@@ -34,10 +34,11 @@ var args struct {
 }
 
 var Cmd = &cobra.Command{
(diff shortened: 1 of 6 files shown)


mrbarge:
@vkareh - wondering if you could please review? Thanks!

vkareh:
This works really well, thanks!
/lgtm

PR #77, opened 2020-05-24
add --wide option to print console url to the list

bmeng:
For some of the auth providers, people need to login to the cluster via the web console.

To print the console url to the list will be a little convenient if we want to login to multiple clusters.

### DIFF
--- a/cmd/ocm/account/status/cmd.go
+++ b/cmd/ocm/account/status/cmd.go
@@ -86,8 +86,8 @@ func run(cmd *cobra.Command, argv []string) error {
 	// Display user and which server they are logged into
 	currAccount := response.Body()
 	currOrg := currAccount.Organization()
-	fmt.Println(fmt.Sprintf("User %s on %s in org '%s' %s (external_id: %s)",
-		currAccount.Username(), cfg.URL, currOrg.Name(), currOrg.ID(), currOrg.ExternalID()))
+	fmt.Printf("User %s on %s in org '%s' %s (external_id: %s)",
+		currAccount.Username(), cfg.URL, currOrg.Name(), currOrg.ID(), currOrg.ExternalID())
 
 	// Display roles currently assigned to the user
 	roleSlice, err := acc_util.GetRolesFromUsers([]*amv1.Account{currAccount}, connection)
--- a/cmd/ocm/cluster/list/list.go
+++ b/cmd/ocm/cluster/list/list.go
@@ -40,6 +40,7 @@ var args struct {
 	step      bool
 	columns   string
 	padding   int
+	wide      bool
 }
 
 var managed bool
@@ -80,6 +81,12 @@ func init() {
 		"padding",
 		-1,
 		"Change all column sizes.",
+	)
+	fs.BoolVar(
+		&args.wide,
+		"wide",
+		false,
+		"Print more information",
 	)
 }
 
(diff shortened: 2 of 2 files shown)


bmeng:
[test]

bmeng:
[test]

bmeng:
[test]

bmeng:
@vkareh Could you help take a look? the ci test always fail with the dependencies fetching.

vkareh:
@bmeng, I'm getting the following lint errors:
```
~/g/s/g/o/ocm-cli (⎇ pull/77) $ golangci-lint run
cmd/ocm/cluster/list/list.go:93:1: cyclomatic complexity 31 of func `run` is high (> 30) (gocyclo)
^
cmd/ocm/cluster/list/list.go:36:10: struct of size 88 bytes could be of size 80 bytes (maligned)
var args struct {
         ^
```

bmeng:
Hi @vkareh, thanks for the debugging suggestion.
I have addressed all the issues and the `golangci-lint` run passed on my local. (However changed the existing code which will need more reviews.)

[10:33:19 bmeng@asrock ~/go/src/github.com/openshift-online/ocm-cli] (wide)$ golangci-lint run 
[10:33:24 bmeng@asrock ~/go/src/github.com/openshift-online/ocm-cli] (wide)$ 

But I still get the error in the ci build log.
Looks like it is an known issue [1] for the version of software which are installed on the jenkins slave. 

[1] https://github.com/golang/go/issues/38373

vkareh:
[test]

bmeng:
[test]

bmeng:
@vkareh the test passed. can you help review the PR?

cmd/ocm/account/status/cmd.go
```diff
@@ -86,8 +86,8 @@ func run(cmd *cobra.Command, argv []string) error {
 	// Display user and which server they are logged into
 	currAccount := response.Body()
 	currOrg := currAccount.Organization()
-	fmt.Println(fmt.Sprintf("User %s on %s in org '%s' %s (external_id: %s)",
-		currAccount.Username(), cfg.URL, currOrg.Name(), currOrg.ID(), currOrg.ExternalID()))
+	fmt.Printf("User %s on %s in org '%s' %s (external_id: %s)",
+		currAccount.Username(), cfg.URL, currOrg.Name(), currOrg.ID(), currOrg.ExternalID())
```
cblecker, line 90: If this is changing from a Println, do we need to add a `\n`?

cmd/ocm/cluster/list/list.go
```diff
@@ -142,6 +148,7 @@ func run(cmd *cobra.Command, argv []string) error {
 
 	size := 100
 	index := 1
+	managed := args.managed
```
cblecker, line 151: I don't think this will do what you think it will. See https://github.com/openshift-online/ocm-cli/pull/81

cmd/ocm/cluster/list/list.go
```diff
@@ -112,24 +117,25 @@ func run(cmd *cobra.Command, argv []string) error {
...
 	var argFilter string
 	if len(argv) == 1 && argv[0] != "" {
 		argFilter = fmt.Sprintf("(name like '%%%s%%' or id like '%%%s%%')", argv[0], argv[0])
 	}
 
 	// Update our column name and padding variables:
+	paddingByColumn := []int{34, 40, 60, 20}
+
+	// Print console url if wide parameter specified
+	if args.wide {
```
cblecker, line 130: This arg should probably conflict with manually specifying columns?

vkareh: This is a good point. Alternatively we might want to allow adding columns to the wide view? The only difference between `wide=true` and `wide=false` is the console URL.

What's the use case for this change? It might help to narrow down some of the changes. You can always do `ocm cluster list --columns=id,name,console.url`, for example

bmeng: yes, the purpose for the change is to show the console url to the output.

Currently, with the google auth enabled for the clusters, we cannot use  `ocm cluster login <cluster_id>` directly, so print the console url should be a little convenient to login to the clusters.

With the `ocm cluster list --columns=id,name,console.url` the console url in the output will be truncated in most of the cases, so we still need to add the arg `--padding 90`. This will make the whole command quite hard to input. And with the `--padding 90` option it will also expand the width for the columns other than console.url which are not needed for such wider.

PR #93, opened 2020-06-19
Implement create user command

igoihman:


cmd/ocm/create/user/cmd.go
```diff
@@ -0,0 +1,144 @@
...
+	group      string
+}
+
+var Cmd = &cobra.Command{
+	Use:     "user",
+	Aliases: []string{"users"},
+	Short:   "Configure user access for cluster",
+	Long:    "Configure user access for cluster",
+	Example: `# Add a user to the dedicated-admins group
+  ocm create user myuser --cluster=mycluster --group=dedicated-admins`,
```
zgalor, line 38: maybe add an example of how to create multiple users

cmd/ocm/create/user/cmd.go
```diff
@@ -0,0 +1,144 @@
...
+		Get().
+		Send()
+	if err != nil {
+		return fmt.Errorf("Group '%s' in cluster '%s' doesn't exist", args.group, clusterKey)
+	}
+
+	for _, username := range strings.Split(users, ",") {
+		user, err := cmv1.NewUser().ID(username).Build()
+		if err != nil {
+			return fmt.Errorf("Failed to create '%s' user '%s' for cluster '%s'", args.group, username, clusterKey)
```
zgalor, line 129: Does this work `Best Effort`?
We create users until we reach a user that we failed.
On first failure, we stop, and all the users that we succeeded are created.

say I had a list of 3 users and the second one already exists.
So the client will see:
```
> ocm create user u1,u2,u3 --cluster=mycluster --group=dedicated-admins`
{
  "kind": "User",
  "href": "/api/clusters_mgmt/v1/clusters/11111/groups/dedicated-admins/users/u1",
  "id": "u1"
}

Failed to add 'dedicated-admins' user 'u2' to cluster 'mycluster': user 'u2' already exists
```

I wonder if continuing would be better user experience. i.e:
```
> ocm create user u1,u2,u3 --cluster=mycluster --group=dedicated-admins`
{
  "kind": "User",
  "href": "/api/clusters_mgmt/v1/clusters/11111/groups/dedicated-admins/users/u1",
  "id": "u1"
}

Failed to add 'dedicated-admins' user 'u2' to cluster 'mycluster': user 'u2' already exists

{
  "kind": "User",
  "href": "/api/clusters_mgmt/v1/clusters/11111/groups/dedicated-admins/users/u3",
  "id": "u3"
}

```

WDYT?

igoihman: makes sense, I will change it

cmd/ocm/create/user/cmd.go
vkareh: This is marked as resolved, but I don't see the example

cmd/ocm/create/user/cmd.go
vkareh: This is marked as resolved but I don't see any change (or any commits after this message)

cmd/ocm/create/user/cmd.go
```diff
@@ -0,0 +1,148 @@
...
+			Group(args.group).
+			Users().
+			Add().
+			Body(user).
+			Send()
+		if err != nil {
+			fmt.Printf("Failed to add '%s' user '%s' to cluster '%s': %v\n", args.group, username, clusterKey, err)
+			continue
+		}
+		fmt.Printf("Failed to add '%s' user '%s' to cluster '%s': %v\n", args.group, username, clusterKey, err)
```
vkareh, line 144: This message seems wrong

cmd/ocm/create/user/cmd.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+		_, err = clusterCollection.Cluster(cluster.ID()).
+			Groups().
+			Group(args.group).
+			Users().
+			Add().
+			Body(user).
+			Send()
+		if err != nil {
+			fmt.Printf("Failed to add '%s' user '%s' to cluster '%s': %v\n", args.group, username, clusterKey, err)
```
vkareh, line 141: This should be `fmt.Errorf`

cmd/ocm/create/user/cmd.go
igoihman: I forgot to edit the sentence after copy-pasting :) Updated now.

cmd/ocm/create/user/cmd.go
vkareh: Based on offline discussion, let's leave this as `fmt.Printf`

PR #95, opened 2020-06-19
Change oauthURL for RHMI

igoihman:


vkareh:
This looks good, but it's also missing from the OpenID provider.

vkareh:
Thanks! /lgtm

vkareh:
pipeline is complaining about using `"rhmi"` too many times. One possible solution is to replace the string twice, regardless of product ID:
```go
oauthURL = strings.Replace(consoleURL, "console-openshift-console", "oauth-openshift", 1)
oauthURL = strings.Replace(oauthURL, "solution-explorer", "oauth-openshift", 1)
```
One of them will always be a no-op but it should work. What do you think?

igoihman:
> pipeline is complaining about using `"rhmi"` too many times. One possible solution is to replace the string twice, regardless of product ID:
> 
> ```go
> oauthURL = strings.Replace(consoleURL, "console-openshift-console", "oauth-openshift", 1)
> oauthURL = strings.Replace(oauthURL, "solution-explorer", "oauth-openshift", 1)
> ```
> 
> One of them will always be a no-op but it should work. What do you think?

I can just use a const for rhmi

vkareh:
another nitpick: since you are now importing a common `cluster` pkg, you can now add to it a function `getOauthURL(...)` and reuse it across all three files

PR #97, opened 2020-06-24
Implement IdP deletion

igoihman:


igoihman:
@vkareh PTAL

vkareh:
/lgtm

But I'm also wondering whether we want to add `idp` to `urls.Resources()` so that bash completion works. It's not a blocker, but would certainly be nice to have completion, but the fact that we are reusing the `delete` command as both a resource versus a curl-type command may cause some issues at some point... maybe we can rethink `pkg/urls` in a separate patch

vkareh:
This seems to work, but it's definitely a hack:
```patch
diff --git a/pkg/urls/url_expander.go b/pkg/urls/url_expander.go
index 03026b7..54b4b57 100644
--- a/pkg/urls/url_expander.go
+++ b/pkg/urls/url_expander.go
@@ -49,6 +49,7 @@ var individualResourceURLs = map[string]string{
        "role":         "/api/accounts_mgmt/v1/roles/",
        "sku":          "/api/accounts_mgmt/v1/skus/",
        "sku_rule":     "/api/accounts_mgmt/v1/sku_rules",
+       "idp":          "idp",
 }
 
 // Expand returns full URI to UHC resources based on an alias. An alias
```
Thoughts?

igoihman:
I wonder if we should just implement it using a different verb, like `dlt` and leave the curel-type commands intact.WDYT?

vkareh:
no, `delete` is a lot more natural, I think...
the `delete` command is just a top-level command/verb and doesn't do anything on its own, so we can leave them both together - maybe have a local function (in the delete/cmd.go file) that concatenates urls.Resources() with the child commands of itself

PR #99, opened 2020-07-03
Implement create ingress

igoihman:


vkareh:
/lgtm

Tested locally with my CS running with https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/1979

cmd/ocm/create/ingress/cmd.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// Get the client for the cluster management api
+	clusterCollection := connection.ClustersMgmt().V1().Clusters()
+
+	cluster, err := c.GetCluster(clusterCollection, clusterKey)
+	if err != nil {
+		return fmt.Errorf("Failed to get cluster '%s': %v", clusterKey, err)
+	}
+
+	if cluster.CloudProvider().ID() != "aws" {
```
vkareh, line 114: This check might be unnecessary if the API already fails it (which it should). However, this error message is much nicer than whatever the API comes up with

igoihman: I added it only in order to get a nicer message

PR #101, opened 2020-07-04
remove new connection boilerplate code

igoihman:


pkg/ocm/connection.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Config sets the configuration that the connection will use to authenticate the user
+func (b *ConnectionBuilder) Config(value *config.Config) *ConnectionBuilder {
+	b.cfg = value
+	return b
+}
+
+// Build uses the information stored in the builder to create a new OCM connection.
+func (b *ConnectionBuilder) Build() (result *sdk.Connection, err error) {
+	if b.cfg == nil {
```
cben, line 44: Subjective, optional: from the way this is called `NewConnection().Build()`, it's not clear a default config file will be read.
I'd rather make that explicit function somehow e.g. `NewConnection().FromConfig().Build()`.  Or even a top-level wrapper like `ConnectionFromConfig()` encapsulating everything including the Build().

The 2nd half checking `Armed()` etc. is fine in Build(), validation is one of the purposes of builder pattern.

tzvatot:
LGTM

PR #102, opened 2020-07-05
Implement list cluster identity providers

igoihman:


igoihman:
@vkareh PTAL

cmd/ocm/list/idp/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+		return "LDAP"
+	case "OpenIDIdentityProvider":
+		return "OpenID"
+	}
+
+	return ""
+}
+
+func getAuthURL(cluster *cmv1.Cluster, idpName string) string {
+	oauthURL := strings.Replace(cluster.Console().URL(), "console-openshift-console", "oauth-openshift", 1)
```
vkareh, line 122: This will be a problem for RHMI clusters. You should also replace "solution-explorer" with "oauth-openshift"

igoihman:
@vkareh PTAL

PR #109, opened 2020-07-24
Implement delete ingress

igoihman:


cmd/ocm/delete/cmd.go
```diff
@@ -45,6 +46,8 @@ func init() {
 	fs := Cmd.Flags()
 	arguments.AddParameterFlag(fs, &args.parameter)
 	arguments.AddHeaderFlag(fs, &args.header)
+	Cmd.AddCommand(idp.Cmd)
```
vkareh, line 49: carryover from #97 ?

igoihman: bad rebase, thanks.

cmd/ocm/delete/ingress/cmd.go
```diff
@@ -0,0 +1,134 @@
...
+	"regexp"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/spf13/cobra"
+
+	c "github.com/openshift-online/ocm-cli/pkg/cluster"
+	"github.com/openshift-online/ocm-cli/pkg/ocm"
+)
+
+var ingressKeyRE = regexp.MustCompile(`^[a-z0-9]{3,5}$`)
```
vkareh, line 27: Since you're not tracking the `api` "ingress", this regexp can be narrowed down to `{4,5}` (btw, thanks for skipping the `api` "ingress", I'm not super thrilled at having mixed it with the ingress and I think it belongs in an `edit cluster` command instead). :+1:

vkareh:
Other than comment about the regexp, this /lgtm

PR #113, opened 2020-07-28
Implement list addons

igoihman:
```
 ./ocm list addons --cluster=irit-test2
ID               NAME             STATE
dba-operator     DBA Operator     not installed
prow-operator    Prow Operator    not installed
```

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -21,8 +21,17 @@ import (
 	"regexp"
 	"strings"
 
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 )
+
+type AddOnItem struct {
+	ID        string
+	Name      string
+	State     string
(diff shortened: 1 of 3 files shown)


vkareh:
```
14:02:55 pkg/cluster/cluster.go:29:6: type name will be used as cluster.ClusterAddOn by other packages, and that stutters; consider calling this AddOn (golint)
14:02:55 type ClusterAddOn struct {
14:02:55      ^
14:02:55 make: *** [lint] Error 1
```
I don't agree with the suggestion of the linter, but we should probably find a better name that the linter doesn't complain about

pkg/cluster/cluster.go
```diff
@@ -21,9 +21,18 @@ import (
 	"regexp"
 	"strings"
 
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 )
 
+type AddOn struct {
```
vkareh, line 29: We should call this `AddOnItem` or `AddOnStatus` or some other name that doesn't get confused with `cmv1.AddOn`

PR #118, opened 2020-08-04
Support git style ocm plugin

wanghaoran1988:


wanghaoran1988:
[test]

cben:
Neat!

Can you document the plugin contract in the README?

If both a builtin command `foo` and a plugin `ocm-foo` exist, which should win?
What does git do?
What does https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/ do?

cben:
Also `ocm help` / `ocm --help` should at least mention that plugins may provide additional commands.  Or list all plugins.

cben:
@vkareh @jeremyeder PTAL.  
If there long-term goal is to to merge moactl & ocm-cli (?), perhaps this is a good route, via a plugin providing `ocm moa ...` commands, or something like that.  No idea if that fits customer experience / mental model you want, just throwing out the idea...
Once the public interface is `ocm foo ...`, it's up to us whether to keep it as separate `ocm-foo` binary or combine into one.

wanghaoran1988:
> If both a builtin command `foo` and a plugin `ocm-foo` exist, which should win?
> What does git do?
> What does https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/ do?

It will use the subcommand instead of the plugin, as it will search the ocm command tree to find it first, if not exists then find the plugin, that's also how kubectl works.

> Also ocm help / ocm --help should at least mention that plugins may provide additional commands. Or list all plugins.
We can also have a command `ocm plugin list` that list all the plugins, that's how kubectl works.

jeremyeder:
Actually the long term goal is to merge moactl into the AWS CLI, once they have their plugin infrastructure.  But that is a ways off.

Regardless, plugins for OCM can proceed however you'd like.  I just don't see ```ocm moa``` as one of the use-cases.

wanghaoran1988:
@cben I updated the README doc, could you take another look?

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,113 @@
...
+	// in order to recognize valid plugin filenames.
+	// The first filepath to match a prefix is returned.
+	Lookup(filename string) (string, bool)
+	// Execute receives an executable's filepath, a slice
+	// of arguments, and a slice of environment variables
+	// to relay to the executable.
+	Execute(executablePath string, cmdArgs, environment []string) error
+}
+
+// DefaultPluginHandler implements PluginHandler
```
cben, line 27:
```suggestion
// DefaultPluginHandler implements Handler
```
also, given that we're inside pkg/plugin, from outside these are accessed as `plugin.DefaultPluginHandler` and `plugin.NewDefaultPluginHandler()`, you could drop Plugin from the name.  not critical.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,113 @@
...
+	if runtime.GOOS == "windows" {
+		// #nosec G204
+		cmd := exec.Command(executablePath, cmdArgs...)
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		cmd.Stdin = os.Stdin
+		cmd.Env = environment
+		err := cmd.Run()
+		if err == nil {
+			os.Exit(0)
```
cben, line 66: I find it a bit confusing that this function _sometimes_ returns an error to the caller and _sometimes_ doesn't return (via exec syscall or Exit here).

- One option is to make the contract "this function doesn't return", and do os.Exit() on all errors.
    - In particular, [cmd.Run](https://golang.org/pkg/os/exec/#Cmd.Run) may return `ExitError` when plugin was launched successfully but exited with non-0 status, say 42.  
      In the UNIX `Exec` code path, the parent of `ocm` gets all exit statuses directly from the process that replaced it.  As currently written, on Windows, this function returns an error and main.go does Exit(1), losing the plugin's exact status.  
      So for this option, you should handle `ExitError` here by exiting with same status.

- Another clean option is to decide this function always returns.
    - This means you can't use exec syscall, always exec.Command that waits for subprocess.  
      IMHO that's good idea anyway for code simplicity and testing :+1:.
    - main.go then should detect ExitError and propogate `ExitError.ExitCode()`.

  P.S. `sycall` package is [frozen & deprecated](https://golang.org/s/go1.4-syscall).  If you want it, [x/sys/unix.Exec](https://godoc.org/golang.org/x/sys/unix#Exec) is preferred.  (though I expect sycall.Exec to work forever...)

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,113 @@
...
+			remainingArgs = remainingArgs[:len(remainingArgs)-1]
+			continue
+		}
+
+		foundBinaryPath = path
+		break
+	}
+
+	if len(foundBinaryPath) == 0 {
+		return nil
```
cben, line 104: I think `return nil` has ambiguous meaning with current contract.  It means "not found, use built-in command" here, but it means "executed plugin (on windows), with no issues" below?

Or did you mean to return a "not found" error here?

cben: One idea for a (major) refactoring is splitting this function report whether it found a plugin, and let main call execute it separately.  But the catch is that it also has to communicate how many arguments matched plugin name / how many to pass to it...  

One "OOP" way to do this cleanly is make an interface that just has an `Execute() error` method, with no args — an object satisfying it should already carry necessary info (executablePath and args) inside.
Then this function becomes one that either returns such an object, or nil if not found (and maybe an error?).
The object .Execute() can be called separately and that can return an error about execution (or print it and os.Exit).

This is just an idea, and probably overkill.  There are simpler changes that fix it...

But please decide what is the contract of this function for following scenarios:
1. no plugin found
2. plugin found, some OS error starting it
3. plugin found, process finished, non-0 exit status
4. plugin found, process finished, 0 exit status.

(1) and (2) may be same to caller, especially under current policy that built-in command wins over plugin of same name.

cmd/ocm/main.go
```diff
@@ -96,7 +97,20 @@ func main() {
...
+	if len(args) > 1 {
+		cmdPathPieces := args[1:]
 
+		// only look for suitable extension executables if
+		// the specified command does not already exist
+		if _, _, err := root.Find(cmdPathPieces); err != nil {
+			if err := plugin.HandlePluginCommand(pluginHandler, cmdPathPieces); err != nil {
+				fmt.Fprintf(os.Stderr, "%v\n", err)
+				os.Exit(1)
+			}
```
cben, line 111: If there was no error, this proceeds to execute built-in command.
in most scenarios, if a plugin was found HandlePluginCommand() will not return at all, but i think on windows it currently _may_ return if plugin run succesfully and returned 0.  

Not saying you should handle this here, just pointing out this scenario is now possible.  See comments on plugin.go on pinning down the contract of HandlePluginCommand()...

pkg/plugin/plugin.go
wanghaoran1988: Hi, see the codes in the main.go also, the general logic here is:
1. Find the command in root cmd tree, if cannot find, then try to find the plugin
2.If the plugin is found, then execute  it, and return what ever error the plugin returned.
3. If the plugin is also not found, we return nil, not error, so the main func will continue to run with the ocm root cmd, and it will return errors cause the cmd not exists.

wanghaoran1988:
@cben  Could you please take another look when you have time?

pkg/plugin/plugin.go
cben: LGTM :+1:  
@igoihman please merge/review.

cmd/ocm/main.go
cben: resolved by last changes :heavy_check_mark:

cben:
LGTM :+1:

@vkareh can you review/merge?

wanghaoran1988:
Kindly ping @igoihman @vkareh

igoihman (approved):
/lgtm

PR #119, opened 2020-08-13
Add tunnel command

cblecker:
This creates a `ocm cluster tunnel` command that will pass the correct flags to the sshuttle tool to create a ssh tunnel to the cluster.


jewzaam:
Is this assuming specific configuration exists on the target cluster?  Thought to moving this out to a plugin later?  Or removing completely once we have backplane?  Wonder if a proper ocm command is necessary vs a script we carry short term given changes coming.

igoihman:
I agree with @jewzaam . `ocm-cli` is a customer facing tool, I don't think this is the right place for introducing this functionality.

cblecker:
Right now this is bridging a gap, @igoihman @jewzaam. The information we need exists in OCM, and we can't access private clusters without these details.

It's possible to have another utility that wraps an `ocm get cluster` or something, but it would be more complex than the above.

This particular command is actually using Cobra's hidden flag, so it won't show up in help texts unless you know about it directly, as it isn't designed for non-RH consumption right now.

As this gap is an active pain point for our current scalability efforts, is it possible to merge this now, and then discuss the long term vision once backplane is in place?

jewzaam:
@igoihman could we move forward with this change if we have a clear agreement on removing this when backplane makes this obsolete?  This is a much cleaner solution from an SREP point of view than a bespoke script for managing the tunnel.  It's not going to be presented in cli help and won't be usable if it is discovered by anybody outside of SREP.

igoihman:
I will review it later today

mbarnes:
I suggested to @cblecker last night to integrate the required bastion jump host into the command.

Currently it's _possible_ to do on the command-line through extra sshuttle arguments, but not very convenient:
```
ocm cluster tunnel $CLUSTERID -- -e "ssh -J $JUMPHOST"
```

Perhaps:
```
ocm cluster tunnel -J/--jumphost=$JUMPHOST $CLUSTERID
```

cblecker:
@mbarnes For simplicity sake, and because this is a temporary command, I think I want to steer away from providing default flag values.. if someone specifies something like `ocm cluster tunnel -J/--jumphost=$JUMPHOST $CLUSTERID -- -e ssh -o StrictHostKeyChecking=no` or some other option, it would break.

cben:
re plugins: #118 will allow `ocm cluster tunnel` to be implemented by a separate `ocm-cluster-tunnel` binary in $PATH, but it doesn't provide any other help other than spaces vs dashes.  So it'd still mean maintaining a separate program (whech can use the SDK).

I don't know enough about backplane to add an opinion.  Irit is sick today.  @vkareh that leaves you to decide.

cmd/ocm/cluster/tunnel/tunnel.go
```diff
@@ -0,0 +1,128 @@
...
+	// Create the client for the OCM API:
+	connection, err := ocm.NewConnection().Build()
+	if err != nil {
+		return fmt.Errorf("failed to create OCM connection: %v", err)
+	}
+	defer connection.Close()
+
+	// Get the client for the resource that manages the collection of clusters:
+	collection := connection.ClustersMgmt().V1().Clusters()
+	clusters, total, err := clusterpkg.FindClusters(collection, args[0], clusterpkg.ClustersPageSize)
```
igoihman, line 66: please validate args[0] exists before calling it

igoihman: we already have `GetCluster()` function in `pkg/cluster/cluster.go`

pkg/cluster/cluster.go
```diff
@@ -206,3 +212,73 @@ func GetClusterAddOns(connection *sdk.Connection, clusterID string) ([]*AddOnIte
 
 	return clusterAddOns, nil
 }
+
+// DoSurvey will ask user to choose one if there are more than one clusters match the query
+func DoSurvey(clusters []*clustersmgmtv1.Cluster) (cluster *clustersmgmtv1.Cluster, err error) {
```
igoihman, line 217: `DoSurvey` function is more related to cli options and not to cluster related functions that exists in this package.
Please consider using `GetCluster` which will give you a more fine grained result and you won't need to use `DoSurvey` function as part of the new command.
Also, `survey.v1` is outdated and we moved to use `v2` in other commands.

pkg/cluster/cluster.go
```diff
@@ -206,3 +212,73 @@ func GetClusterAddOns(connection *sdk.Connection, clusterID string) ([]*AddOnIte
...
+		return nil
+	})
+	return cluster, err
+}
+
+// FindClusters finds the clusters that match the given key. A cluster matches the key if its
+// identifier is that key, or if its name starts with that key. For example, the key `prd-2305`
+// doesn't match a cluster directly because it isn't a valid identifier, but it matches all clusters
+// whose names start with `prd-2305`.
+func FindClusters(collection *clustersmgmtv1.ClustersClient, key string,
```
igoihman, line 250: I less prefer using this function for finding clusters. It returns a list of clusters that match a key while in my opinion a more smooth experience will be to return just one cluster based on a name or ID.

pkg/cluster/cluster.go
cblecker: These functions were moved verbatim from the existing ocm cluster login, so that the two commands share these code paths.

cblecker: Please see https://github.com/openshift-online/ocm-cli/pull/119/commits/60af13a2f2a99481dfb2f18bacd8b41886ddccb4 where these were just moved out from `ocm cluster login` so they could be reused in `ocm cluster tunnel`.

igoihman: Can you use the `FindCluster` function instead?

igoihman: https://github.com/openshift-online/ocm-cli/blob/master/pkg/cluster/cluster.go#L44

cmd/ocm/tunnel/cmd.go
```diff
@@ -0,0 +1,132 @@
...
+	clusterKey string
+}
+
+var Cmd = &cobra.Command{
+	Use:   "tunnel --cluster=<CLUSTERID|CLUSTER_NAME|CLUSTER_NAME_SEARCH> -- [sshuttle arguments]",
+	Short: "tunnel to a cluster",
+	Long: "Use sshuttle to create a ssh tunnel to a cluster by ID or Name or" +
+		"cluster name search string according to the api: " +
+		"https://api.openshift.com/#/clusters/get_api_clusters_mgmt_v1_clusters",
+	Example: " ocm cluster tunnel <id>\n ocm cluster tunnel %test%",
```
igoihman, line 41: should be `ocm tunnel --cluster=<id>`

igoihman:
It works as expected. I will merge it as is and push a fix to the help message in order to unblock it

igoihman (approved):
/lgtm

PR #120, opened 2020-08-16
Add more options to create cluster command

igoihman:


cmd/ocm/create/cluster/cmd.go
```diff
@@ -39,6 +41,17 @@ var args struct {
 	expirationTime    string
 	expirationSeconds time.Duration
```
vkareh, line 42: Can you make the expiration ones hidden, as we don't support these flags in production?

igoihman:
@vkareh Thanks, I will create another PR for that

PR #121, opened 2020-08-17
Display provision shard name in describe cluster

fahlmant:
Adds the provision shard to `ocm describe cluster`

### DIFF
--- a/pkg/cluster/describe.go
+++ b/pkg/cluster/describe.go
@@ -78,6 +78,17 @@ func PrintClusterDesctipion(connection *sdk.Connection, cluster *cmv1.Cluster) e
 		creator = "N/A"
 	}
 
+	// Find the details of the shard
+	shardPath, err := connection.ClustersMgmt().V1().Clusters().
+		Cluster(cluster.ID()).
+		ProvisionShard().
+		Get().
+		Send()
+	var shard string
+	if shardPath != nil && err == nil {
+		shard = shardPath.Body().HiveConfig().Server()
+	}
+
 	// Print short cluster description:
 	fmt.Printf("\n"+
 		"ID:            %s\n"+
@@ -118,6 +129,9 @@ func PrintClusterDesctipion(connection *sdk.Connection, cluster *cmv1.Cluster) e
 		cluster.CreationTimestamp().Round(time.Second).Format(time.RFC3339Nano),
 		cluster.ExpirationTimestamp().Round(time.Second).Format(time.RFC3339Nano),
 	)
+	if shard != "" {
+		fmt.Printf("Shard:         %v\n", shard)
+	}
 	fmt.Println()
 
 	return nil


pkg/cluster/describe.go
```diff
@@ -97,7 +97,8 @@ func PrintClusterDesctipion(connection *sdk.Connection, cluster *cmv1.Cluster) e
 		"Cluster Admin: %t\n"+
 		"Creator:       %s\n"+
 		"Created:       %v\n"+
-		"Expiration:    %v\n",
+		"Expiration:    %v\n"+
+		"Shard:         %v\n",
```
igoihman, line 101: The shard details and its actual existence should be hidden for customers as they are not authorized to view it.

pkg/cluster/describe.go
```diff
@@ -118,6 +118,10 @@ func PrintClusterDesctipion(connection *sdk.Connection, cluster *cmv1.Cluster) e
 		cluster.CreationTimestamp().Round(time.Second).Format(time.RFC3339Nano),
 		cluster.ExpirationTimestamp().Round(time.Second).Format(time.RFC3339Nano),
 	)
+	shard := cluster.ProvisionShard()
```
vkareh, line 121: I don't think this is going to work. ProvisionShards don't get populated by the OCM API as part of the cluster object, but need a separate request (i.e. if you do `ocm get /api/.../cluster/{id}` you do not get the provision shard, you need to query the `/cluster/{id}/provision_shard` endpoint explicitly)

There are two ways to approach this:
1. Do a separate client query in `ocm describe` - this is easier and can be done as part of this PR with minimal changes
2. Expose provision shard in the cluster object directly (only populating it given the correct auth) - this is harder and will require a re-deploy of OCM, but will allow the usage of `ocm list clusters --columns=provision_shard.hive_config.server` without additional changest to `ocm-cli`

jharrington22: Thanks @vkareh.

Its seems odd to me that the ocm-cli describe and the API would return different information that what is in the cluster object but it seems that we are already doing this for subscription and the cluster author. I'm +1 to just adding another call [here](https://github.com/openshift-online/ocm-cli/blob/1f8511a2e9a6ee67c01d9a3c110614a94b0fbbdb/pkg/cluster/describe.go#L28) to query for the shard ID if the user authorized. This would be quick and allow us to move forward.

fahlmant: @vkareh I'm going to give option 1 a try. Seems like the best path forward here

pkg/cluster/describe.go
```diff
@@ -78,6 +78,19 @@ func PrintClusterDesctipion(connection *sdk.Connection, cluster *cmv1.Cluster) e
...
 
+	// Find the details of the shard
+	shardPath, err := connection.ClustersMgmt().V1().Clusters().
+		Cluster(cluster.ID()).
+		ProvisionShard().
+		Get().
+		Send()
+	var shard string
+	if shardPath == nil || err != nil {
+		shard = "N/A"
```
vkareh, line 89: This exposes knowledge about shards to users without authorization. I would still leave as a conditional print (only print it out if it's not `nil`)

fahlmant: :+1:

fahlmant: Done

igoihman:
LGTM

fahlmant:
@vkareh Done

PR #124, opened 2020-08-19
Assume expiration is 0 when missing in the token claim

tzvatot:
@igoihman @cben PTAL


igoihman:
please also update the version in `pkg/info/info.go`

tzvatot:
> please also update the version in `pkg/info/info.go`

Done. PTAL

igoihman (approved):
/lgtm

PR #127, opened 2020-08-24
Remove RHMI specific implementation

igoihman:


igoihman:
@vkareh PTAL

pkg/cluster/cluster.go
```diff
@@ -204,14 +204,9 @@ func CreateCluster(cmv1Client *cmv1.Client, config Spec) (*cmv1.Cluster, error)
 }
 
 func GetClusterOauthURL(cluster *cmv1.Cluster) string {
-	var oauthURL string
 	consoleURL := cluster.Console().URL()
-	if cluster.Product().ID() == "rhmi" {
-		oauthURL = strings.Replace(consoleURL, "solution-explorer", "oauth-openshift", 1)
```
vkareh, line 210: hmmm... This is a tricky one, we currently have RHMI clusters in production. They are managed internally AFAIK, but I wouldn't remove this logic just yet until we can transition them to OSD+3scale (or whatever the combo will be) or until we get confirmation from the integr8ly team that we no longer need the Solution Explorer as the main entry point for clusters.

PR #129, opened 2020-08-30
Add channel group describe cluster

nimrodshn:
Example output:
```
ID:            1J3WlGOQJolyPguJ5M3vIOF8TAh
External ID:   
Name:          abc.act2.autom8.in
API URL:       
API Listening: 
Console URL:   
Masters:       3
Infra:         0
Computes:      4
Product:       osd
Provider:      aws
Region:        eu-west-1
Multi-az:      false
CCS:           false
Channel Group: stable
Cluster Admin: false
Creator:       jhernand-rh
Created:       2019-03-27T22:06:51Z
Expiration:    0001-01-01T00:00:00Z
Shard:         https://api.hive-stage.openshift.com:443
```

cc: @igoihman @cben @tzvatot  ptal.


pkg/cluster/describe.go
```diff
@@ -89,6 +89,16 @@ func PrintClusterDesctipion(connection *sdk.Connection, cluster *cmv1.Cluster) e
 		shard = shardPath.Body().HiveConfig().Server()
 	}
 
+	var channelGroup string
+	version, err := connection.ClustersMgmt().V1().
```
igoihman, line 93: while this is technically correct implementation, we should have the channel group as a cluster resource attribute and the information should be derived from there.

nimrodshn: Sure, I will close this PR and rewrite it once the channel group is added as an attribute to the Cluster resource.

PR #133, opened 2020-09-03
Support creating CCS clusters

igoihman:


igoihman:
@vkareh PTAL

nimrodshn:
@igoihman LGTM :+1: 

We should move the AWS struct under CCS in a follow up work.

PR #135, opened 2020-09-04
new sub-command to show the plugins

bmeng:


cmd/ocm/main.go
```diff
@@ -38,13 +39,14 @@ import (
...
 	"github.com/openshift-online/ocm-cli/cmd/ocm/patch"
+	plugincmd "github.com/openshift-online/ocm-cli/cmd/ocm/plugin"
 	"github.com/openshift-online/ocm-cli/cmd/ocm/post"
 	"github.com/openshift-online/ocm-cli/cmd/ocm/token"
 	"github.com/openshift-online/ocm-cli/cmd/ocm/tunnel"
 	"github.com/openshift-online/ocm-cli/cmd/ocm/version"
 	"github.com/openshift-online/ocm-cli/cmd/ocm/whoami"
 	"github.com/openshift-online/ocm-cli/pkg/arguments"
-	"github.com/openshift-online/ocm-cli/pkg/plugin"
+	plugin "github.com/openshift-online/ocm-cli/pkg/plugin"
```
cben, line 49: cosmetic: consider renaming the directory to pkg/plugins so we won't need import renaming.  your call.

cben: actually ignore that.

@igoihman @vkareh Do you have an opinion on the command naming itself?  `ocm plugin list` vs `ocm plugins list` vs `ocm list plugins` (may be confusing because "ocm list foo" is for API objects not local stuff) vs just `ocm plugins`?
I recall somebody had an opinion on verb/noun order but can't find it...

cmd/ocm/plugin/list/cmd.go
```diff
@@ -0,0 +1,139 @@
...
+// uniquePath remove the duplicate items from the PATH
+func uniquePath(path []string) []string {
+	keys := make(map[string]int)
+	uniPath := make([]string, 0)
+
+	for _, p := range path {
+		keys[p] = 1
+	}
+
+	for element := range keys {
```
cben, line 110: this permutes the order of $PATH (to whatever order the hash table has), but since we only care about _set_ of plugins, that's OK :heavy_check_mark:

cmd/ocm/main.go
vkareh: yes, regarding API resources, we're trying to standardize on `ocm {verb} {noun}`, which would make this `ocm list plugins`, but you bring up a good point: this is not an API resource, so maybe it's ok to break convention??

Also, cobra supports aliases, so I'd make the default to be plural `plugins` (since it's a list of many) with an alias to `plugin`.

cmd/ocm/plugin/list/cmd.go
```diff
@@ -0,0 +1,139 @@
...
+			}
+
+			if !strings.HasPrefix(f.Name(), pluginPrefix) {
+				continue
+			}
+
+			plugin := f.Name()
+
+			var pluginOutput []string
+			absPath := dir + "/" + plugin
```
cben, line 83: `filepath.Join(dir, plugin)` may be more correct in corner cases, specifically on windows.

cmd/ocm/plugin/list/cmd.go
```diff
@@ -0,0 +1,139 @@
...
+	newPath := uniquePath(defaultPath)
+	pluginPrefix := "ocm-"
+
+	columns := "NAME,PATH"
+	paddingByColumn := []int{20, 30}
+
+	table.PrintPadded(os.Stdout, strings.Split(columns, ","), paddingByColumn)
+
+	for _, dir := range newPath {
+		if _, err := os.Stat(dir); os.IsNotExist(err) {
```
cben, line 64: +1 to continuing after non-existant PATH entries, with or without warning.  
Most programs, including `which` ignore them silently.

$PATH may contain empty entries which mean current working dir, so it's better to first do:

```
if dir == "" {
  dir = "."
}
```

> A zero-length prefix is a legacy feature that indicates the current working directory. It appears as two adjacent <colon> characters ( "::" ), as an initial <colon> preceding the rest of the list, or as a trailing <colon> following the rest of the list. A strictly conforming application shall use an actual pathname (such as .) to represent the current working directory in PATH.  
-- [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03)

cben:
Small nits but overall this looks good.
Consider moving part of the logic from the cmd into pkg/plugin.

cben:
Crazy idea: can we make the plugins appear as part of regular `ocm --help`?

bmeng:
> Crazy idea: can we make the plugins appear as part of regular `ocm --help`?

Currently, it will show the `plugin` subcommand in `ocm --help`. Do you mean we should hide this?

igoihman:
which plugins will the user see?I am against going to a direction where backplane will be visible to customers

bmeng:
@cben @vkareh updated a bit based on the comments, help review again?
for the singular and plural of the `plugin`, since all the other sub-commands are all using singular, so I also leave this as the same, just adding the alias for the plural.

cben:
> > Crazy idea: can we make the plugins appear as part of regular ocm --help?

> Currently, it will show the plugin subcommand in ocm --help. Do you mean we should hide this?

I meant include the output you're producing here somewhere somewhere inside ocm help output.  And then we might not need a separate command.
But I suppose consistency to kubectl is best goal here, so functionality LGTM :100:
Will re-review code in a moment...

## kubectl
```
$ kubectl --help
...
Other Commands:
  ...
  plugin        Provides utilities for interacting with plugins.
...

$ kubectl plugin
Provides utilities for interacting with plugins.

 Plugins provide extended functionality that is not part of the major command-line distribution. Please refer to the
documentation and examples for more information about how write your own plugins.

 The easiest way to discover and install plugins is via the kubernetes sub-project krew. To install krew, visit
https://github.com/kubernetes-sigs/krew/#installation

Available Commands:
  list        list all visible plugin executables on a user's PATH

Usage:
  kubectl plugin [flags] [options]

Use "kubectl <command> --help" for more information about a given command.
Use "kubectl options" for a list of global command-line options (applies to all commands).
```

```
$ kubectl plugin list
The following compatible plugins are available:

/home/cben/.krew/bin/kubectl-krew
/home/cben/.krew/bin/kubectl-ns
/home/cben/.krew/bin/kubectl-tree
```

## This PR

```diff
 $ ./ocm --help
 ...
 Available Commands:
   account     Get information about users.
   cluster     Get information about clusters
   completion  Generates bash completion scripts
   config      get or set configuration variables
   create      Create a resource from stdin
   delete      Send a DELETE request
   describe    Show details of a specific resource
   edit        Edit a resource from stdin
   get         Send a GET request
   help        Help about any command
   list        List all resources of a specific type
   login       Log in
   logout      Log out
   patch       Send a PATCH request
+  plugin      Get information about plugins
   post        Send a POST request
   token       Generates a token
   version     Prints the version
   whoami      Prints user information
...
```

```
$ ./ocm plugin
Get information about installed ocm plugins

Usage:
  ocm plugin [command]

Aliases:
  plugin, plugins

Available Commands:
  list        list ocm plugins

Flags:
  -h, --help   help for plugin

Global Flags:
      --alsologtostderr                  log to standard error as well as files
      --debug                            Enable debug mode.
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory
      --logtostderr                      log to standard error instead of files (default true)
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
  -v, --v Level                          log level for V logs
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging

Use "ocm plugin [command] --help" for more information about a command.
```

```
$ ./ocm plugin list
NAME                PATH                          
ocm-service-log     /home/cben/go/bin             
```
(ocm-service-log is not intended as a plugin, just happens to start with ocm- but Working As Intended :)

cmd/ocm/plugin/list/cmd.go
```diff
@@ -0,0 +1,147 @@
...
+			continue
+		}
+
+		if dir == "" {
```
cben, line 69: This didn't work:
```
$ echo '#!/bin/nl' > ./ocm-foo; chmod +x ./ocm-foo
$ env PATH=":/" ./ocm foo  # Linux does find the executable
     1	#!/bin/nl
$ env PATH=":/" ./ocm plugin list  # this doesn't
NAME                PATH                          
$
```

The reason is `os.IsNotExist("")` above already led to `continue`.  
You need to move swap the order, or better yet move this normalization into `uniquePath()` to avoid double printing for (admittedly silly ;-) values that contain both "" and "." e.g. `PATH="/bin::.:"`.

cben (commented):
One last nit but otherwise great!

bmeng:
```
$ echo '#!/bin/nl' > ./ocm-foo; chmod +x ./ocm-foo
$ env PATH=":::/" ./ocm plugin list
NAME                PATH
ocm-foo             .
```

@cben thanks for the thoroughly review.
Move the piece of blank path handling to the func `uniquePath` and it works as expected.

cben:
@vkareh @igoihman this is ready for merge :+1:

bmeng:
:tada: Thanks!

PR #140, opened 2020-09-10
Use `--token` as argument for `cluster login`

mmazur:
Currently it's `ocm cluster login --browser`, but the page that redirects to displays the API token, hence `--token` feels more fitting.

Also, in practice this will be mostly used to log in via CLI, so `--browser` is sort of the opposite of what one would use it for.


igoihman:
LGTM. @NautiluX FYI

PR #146, opened 2020-09-17
Implement Edit cluster command

igoihman:


igoihman:
@vkareh PTAL

vkareh:
very nice

/lgtm

PR #147, opened 2020-09-21
Add token generation command

rawsyntax:


cmd/ocm/token/cmd.go
```diff
@@ -111,10 +123,21 @@ func run(cmd *cobra.Command, argv []string) error {
...
-	// Get the tokens:
-	accessToken, refreshToken, err := connection.Tokens()
-	if err != nil {
-		return fmt.Errorf("Can't get token: %v", err)
+	var accessToken string
+	var refreshToken string
+
+	if args.generate {
+		// Get new tokens:
+		accessToken, refreshToken, err = connection.Tokens(15 * time.Minute)
```
vkareh, line 131: would it make sense to make this configurable?
```sh
ocm token --generate # 15 minutes
ocm token --generate=10 # 10 minutes
# etc...
```
maybe not, but wondering, since it's being hard-coded

rawsyntax: There is an upper limit so I didn't want to allow the user to supply nonsensical values like say 60 minutes

vkareh: makes sense

PR #151, opened 2020-09-29
use the cluster id directly without --cluster

bmeng:
if the cluster id is the only parameter for the command, I think it is better to pass the cluster_id directly without `--cluster` for easy use.

@cblecker Please help take a look.

### DIFF
--- a/cmd/ocm/tunnel/cmd.go
+++ b/cmd/ocm/tunnel/cmd.go
@@ -34,35 +34,34 @@ var args struct {
 }
 
 var Cmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


cblecker:
I originally had it like this, but @igoihman preferred the flag format I believe. She can comment :)

igoihman:
I don't have a strong opinion against it, we can drop the `--cluster` flag.
@bmeng CI is complaining about: ```cmd/ocm/tunnel/cmd.go:32:5: `args` is unused (deadcode)```

bmeng:
@igoihman removed the unused var. Please help review again. Thanks.

Issue #67, opened 2020-03-19
CI for this repo?


cblecker:
Should we look into hooking up prow CI for this repo?

vkareh:
that'd be awesome for automatically tagging reviewers and getting auto-merge!

vkareh:
Example configuration for reference (this is what was done for moactl):
* https://github.com/openshift/release/pull/9633/files
* https://github.com/openshift/moactl/pull/25/files

cben:
Huh, turns out we're connected to a CI, a Jenkins maintained by app-sre.  
pr_check.sh file is there for a long time, so I guess the CI has worked for a long time, I just didn't realize it.
Example run: https://ci.ext.devshift.net/job/openshift-online-ocm-cli-gh-pr-check/304/console

however, the CI output is not public-accessible.  Logging in with an unrelated github user I get "Access Denied
anpc is missing the Overall/Read permission".  This is bad for public project, external contributors should be able to see CI output for their PRs.  I'll ask app-sre about that.

jhernand:
CI now runs in GitHub actions. Both the definition of jobs and the logs are accessible to anyone.

PR #155, opened 2020-10-07
Upgrade to golangci-lint 1.31.0

cben:
- I believe most of us have an old golangci-lint 1.10.2 in our PATH, to match what our internal Jenkins has for internal OCM projects.  
  Running from container will allow each repo to control golangci-lint version independently.

- ~~Added `make check` for everything worth running before PR submission.
  Poor man's CI (#67) :grin:~~

If we like these, I can make similar PRs to sdk, moactl etc...


cben:
oh, I didn't realize we already have a pr_check.sh wired to CI :tada:
debugging the failure...

cben:
This is ready.

Makefile
```diff
@@ -21,6 +21,9 @@ export GOPROXY=https://proxy.golang.org
 # Disable CGO so that we always generate static binaries:
 export CGO_ENABLED=0
 
+# Allow overriding: `make lint container_runner=podman`.
+container_runner:=docker
```
igoihman, line 25: can we use podman instead?

cben: It gave some problem, probably too old in ci-ext Jenkins. Let me try again...

cben:
Traced this to https://access.redhat.com/solutions/5106161, ci.ext has has kernel 3.10.0-1062.9.1.el7.x86_64.  Asked app-sre if it's possible to upgrade to kernel-3.10.0-1127.8.2.el7.x86_64.

But meanwhile reverted to docker, podman won't work in CI.

I you want, I can set the Makefile default to podman and override to docker in pr_check.sh.

cben:
Kernel upgraded today, now defaults to podman and works :tada:.

PR #159, opened 2020-10-12
ocm list providers

cben:
Not very useful by itself, but for completeness.

```
$ ./ocm list providers
NAME         DISPLAY NAME
aws          AWS
azure        Azure
baremetal    Bare Metal
gcp          GCP
libvirt      Libvirt
openstack    OpenStack
vsphere      VSphere
```


cben:
Renamed from `cloud_providers` to `providers` for consistency with `ocm create cluster --provider=...`.

cben:
Yeah, I'm working on several of these, but wasn't sure if this one is actually useful.  I'm fine with shelving it.

PR #161, opened 2020-10-13
Improve help and positional arg enforcement in most commands

cben:
Most commands taking positional arguments didn't document them.  Now they do.
Also more consistently using [Cobra conventions for `Use:` line](https://github.com/spf13/cobra/blob/b97b5ead31f7d34f764ac8666e40c214bb8e06dc/command.go#L39-L48), such as "FOO_BAR" rather than "\<foo bar\>" to describe a single parameter.
Chosen approach explained in https://github.com/spf13/cobra/issues/571#issuecomment-707231598

**Behavior change:** Many commands silently ignored unexpected positional arguments (e.g. `ocm account status foo`, `ocm list clusters mycluster foo`), now they'll print an error.

- I didn't touch the generic REST commands such as `get`, `delete` etc.
  - I'm a bit confused by them (they're using `urls.Expand()` to support `get URL` & `get KIND ID` forms, but they _also_ have some Cobra sub-commands).
  - Also there is subtle terminology distinction there of "RESOURCE" (collection/single) vs "PATH" (single) that is not very clear but I wasn't sure how to improve...

### DIFF
--- a/cmd/ocm/account/roles/cmd.go
+++ b/cmd/ocm/account/roles/cmd.go
@@ -32,10 +32,16 @@ var args struct {
 }
 
 var Cmd = &cobra.Command{
-	Use:   "roles [role-name]",
+	Use:   "roles [flags] [ROLE_NAME]",
 	Short: "Retrieve information of the different roles",
-	Long:  "Get description of a role or list of all roles ",
-	RunE:  run,
+	Long:  "Get description of a role or list of all roles",
+	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) > 1 {
+			return fmt.Errorf("Accepts at most 1 role name")
+		}
+		return nil
+	},
+	RunE: run,
 }
 
 func init() {
--- a/cmd/ocm/list/machinepool/cmd.go
+++ b/cmd/ocm/list/machinepool/cmd.go
@@ -52,7 +52,7 @@ func init() {
 		"cluster",
 		"c",
 		"",
-		"Name or ID of the cluster to list the machine pools of (required).",
+		"Name or ID or external_id of the cluster to list the machine pools of (required).",
 	)
 	//nolint:gosec
 	Cmd.MarkFlagRequired("cluster")
--- a/cmd/ocm/list/machinepool/cmd.go
+++ b/cmd/ocm/list/machinepool/cmd.go
@@ -34,7 +34,7 @@ var args struct {
 }
 
 var Cmd = &cobra.Command{
-	Use:     "machinepools",
+	Use:     "machinepools --cluster={NAME|ID|EXTERNAL_ID}",
 	Aliases: []string{"machine-pool", "machine-pools", "machinepool"},
 	Short:   "List cluster machine pools",
 	Long:    "List machine pools for a cluster.",
--- a/cmd/ocm/list/machinepool/cmd.go
+++ b/cmd/ocm/list/machinepool/cmd.go
@@ -40,6 +40,7 @@ var Cmd = &cobra.Command{
 	Long:    "List machine pools for a cluster.",
 	Example: `  # List all machine pools on a cluster named "mycluster"
   ocm list machine-pools --cluster=mycluster`,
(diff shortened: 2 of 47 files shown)


cben:
The `cobra` bump is probably unnecessary, I'm not aware of any concrete changes in cobra this depends on, but since I went diving into current cobra code and re-tested many of these commands with my changes, it made sense to use latest release...

cmd/ocm/list/machinepool/cmd.go
```diff
@@ -40,6 +40,7 @@ var Cmd = &cobra.Command{
 	Long:    "List machine pools for a cluster.",
 	Example: `  # List all machine pools on a cluster named "mycluster"
   ocm list machine-pools --cluster=mycluster`,
+	Args: cobra.NoArgs,
```
ciaranRoche, line 43: Do we need to add to the `Use: "machinepools --cluster={NAME|ID|EXTERNAL_ID}"`

ciaranRoche (commented):
Just one comment inline not sure if its valid, verified some of the commands locally, looks good :+1:

cmd/ocm/list/machinepool/cmd.go
cben: Thanks!  fixed.

ciaranRoche (commented):
/lgtm

cben:
@igoihman @vkareh please review.

cmd/ocm/account/roles/cmd.go
```diff
@@ -32,10 +32,16 @@ var args struct {
 }
 
 var Cmd = &cobra.Command{
-	Use:   "roles [role-name]",
+	Use:   "roles [flags] [ROLE_NAME]",
 	Short: "Retrieve information of the different roles",
-	Long:  "Get description of a role or list of all roles ",
-	RunE:  run,
+	Long:  "Get description of a role or list of all roles",
+	Args: func(cmd *cobra.Command, args []string) error {
```
vkareh, line 38: You could use `cobra.MaximumNArgs(1)` here.

vkareh (approved):
This looks very nice, thanks!

One reason why I haven't used the `cobra.*Args` functions is because the error messages are not particularly helpful, but now that I see this patch I'm wondering if we could wrap the `cobra` functions in such a way that we still defer to them for the validation, but that the error message aligns with the purpose of the command/arg. This could be a later improvement.

cmd/ocm/account/roles/cmd.go
cben: For now I figured custom validation lets me have better control of messages.  
I dumped my thoughts on this https://github.com/spf13/cobra/issues/571#issuecomment-707231598

Part of the problem using built-in validation for us is that we disabled printing usage on error (`SilenceUsage: true`).
The default behavior of dumping full description of all args on any error is too annoying for my taste too.  
I think best behavior would be to print just the `Usage:  ocm account roles [flags] [ROLE_NAME]` line on errors (user can always --help for details); if we had that it'd be reasonable to go back to `cobra.MaximumNArgs` and friends.
Opened https://github.com/spf13/cobra/issues/1252, currently we can't `SetUsageTemplate` to something briefer without breaking --help, working on upstream improvement.

cben: I did add/leave many cobra.*Args and similar, without reviewing each message, largely because of laziness — this PR actually grew from 1 command with wrong help I wanted to fix :rofl:

cben:
@igoihman @vkareh so, can we merge this?  I touched so many files this is likely to grow merge conflicts...

cmd/ocm/account/roles/cmd.go
vkareh: Read your threads upstream and I agree, if we can expose `UseLine()` when Arg fails validation that would get us very close to full-lazy cmd definitions :)

vkareh:
@cben yes, I'm comfortable merging this. Thanks for these improvements :+1: 

/lgtm

PR #166, opened 2020-10-25
feat(tunnel): add a shortening output flag

georgettica:
this was written 100% on the browser, so will pull locally and test after.

the naming of the flag can be changed but I wanted to give the idea of what I wanted
also, it's possible to remove flag altogether and use the string instead

### DIFF
--- a/cmd/ocm/tunnel/cmd.go
+++ b/cmd/ocm/tunnel/cmd.go
@@ -29,6 +29,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var args struct {
+	short bool
+}
+
 var Cmd = &cobra.Command{
 	Use:   "tunnel [flags] {CLUSTERID|CLUSTER_NAME|CLUSTER_NAME_SEARCH} -- [sshuttle arguments]",
 	Short: "tunnel to a cluster",
@@ -42,6 +46,9 @@ var Cmd = &cobra.Command{
 }
 
 func init() {
+	flags := Cmd.Flags()
+
+	flags.BoolVar(&args.short, "short",false, "to shorten the 'sshuttle' path")
 }
 
 func run(cmd *cobra.Command, argv []string) error {
@@ -100,6 +107,11 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 	sshuttleArgs = append(sshuttleArgs, argv[1:]...)
 
+
+	if args.short {
+		path = "sshuttle"
+	}
+	
 	// Output sshuttle command execution string for review
 	fmt.Printf("\n# %s %s\n\n", path, strings.Join(sshuttleArgs, " "))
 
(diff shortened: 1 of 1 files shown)


cmd/ocm/tunnel/cmd.go
```diff
@@ -42,6 +46,9 @@ var Cmd = &cobra.Command{
 }
 
 func init() {
+	flags := Cmd.Flags()
+
+	flags.BoolVar(&args.short, "short",false, "to shorten the 'sshuttle' path")
```
cben, line 51:
```suggestion
	flags.BoolVar(&args.short, "short", false, "to shorten the 'sshuttle' path")
```
to continue the coding-in-browser streak :wink:  I think this will fix the lint error.

cmd/ocm/tunnel/cmd.go
```diff
@@ -100,6 +107,11 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 	sshuttleArgs = append(sshuttleArgs, argv[1:]...)
 
+
+	if args.short {
+		path = "sshuttle"
```
cben, line 112: The resulting command should still work because `LookPath("sshuttle")` found it on PATH.

Q: do you need both modes?  Sounds like you could always print short `sshuttle ...`.
But I never use this, I don't have an opinion here.

@cblecker @drewandersonnz please review.

drewandersonnz: +1 to requiring `sshuttle` in PATH. 

I agree, I don't think a `args.short` should be necessary, I feel like we should rely on path lookup to support multiple OS's.

I believe the intention behind this is because some people use `ocm` via https://github.com/openshift/ocm-container (a Fedora linux environment in a container), yet need to use `sshuttle` on macos, which is found in a different path. This makes copying and pasting the resulting command (which is printed a few lines lower) just that little bit more difficult. 

So to finish, +1 to removing the `short` flag and simply have `path = "sshuttle"`.

You could also just change the print: 
```
	// Output sshuttle command execution string for review
	fmt.Printf("\n# %s %s\n\n", "sshuttle", strings.Join(sshuttleArgs, " "))
```

georgettica: I just didn't want to remove the functionality completely.. but if you are ok with it, so am I.

georgettica:
@cben maybe a CI step that lints and fixes?

georgettica:
now it's done on my laptop, but still needs more 👀  reassigning @drewandersonnz and @cben

cben:
cmd/ocm/tunnel/cmd.go:66:2: ineffectual assignment to `path` (ineffassign)
```
	path, err := exec.LookPath("sshuttle")
	^
```

georgettica:
@cben is it okay now?

georgettica:
let's close is now :)

cben:
O sorry, i had completely missed this.  Dont be afraid to nag :)

LGTM - and more importantly, matcnes drewandersonnz's opinion - :+1: if you want to reopen.

georgettica:
If this is the case I will reshape this and talk to @drewandersonnz on this once it's ready

PR #169, opened 2020-10-27
fix version check when creating a cluster

igoihman:


vkareh:
ouch, I think I need to do this for `rosa` as well, it's been an issue for a while :/

/lgtm

PR #170, opened 2020-11-03
Update integration env default url

wanghaoran1988:


oourfali:
lgtm.

igoihman (approved):

oourfali:
I guess we should revert that, as the environment isn't ready yet for the CLI to consume?

PR #174, opened 2020-11-12
Support creating automatic upgrade policies

igoihman:


igoihman:
@vkareh PTAL

cmd/ocm/create/upgradepolicy/cmd.go
```diff
@@ -81,85 +83,154 @@ func run(cmd *cobra.Command, argv []string) error {
...
+		if err != nil {
+			return fmt.Errorf("Failed to get a valid hour")
+		}
+
+		hourInt, err := strconv.Atoi(strings.Split(hour, ":")[0])
+		if err != nil {
+			return nil
+		}
+
+		cronExpression := fmt.Sprintf("0 %d * * %d", hourInt, dayInt)
```
tzvatot, line 150: The cron expression should be in UTC, while it seems that user might thinks the selection is on local time. This creates some deviation.

PR #179, opened 2020-11-13
Add list orgs command

igoihman:


cmd/ocm/list/org/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+var args struct {
+	columns   string
+	parameter []string
+	padding   int
+}
+
+// Cmd ...
+var Cmd = &cobra.Command{
+	Use:     "organization",
+	Aliases: []string{"organization", "org", "organizations"},
```
cben, line 43: Maybe add plural `orgs` too for consistency.

cmd/ocm/list/org/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+
+func init() {
+	// Add flags to rootCmd:
+	fs := Cmd.Flags()
+	arguments.AddParameterFlag(fs, &args.parameter)
+	fs.StringVar(
+		&args.columns,
+		"columns",
+		"id,name", // Default value gets assigned later as connection is needed.
+		"Organization identifier. Defaults to the organization of the current user.",
```
cben, line 58: description doesn't match the flag.

cmd/ocm/list/org/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+
+		// Fetch next page
+		orgList, err := request.Send()
+		if err != nil {
+			return fmt.Errorf("Failed to retrieve organization list: %v", err)
+		}
+
+		// Display organization information
+		orgList.Items().Each(func(org *amv1.Organization) bool {
+			// String to output marshal -
```
cben, line 113: This function interleaves 2 responsibilities: iterating over pages & printing an item accourding to args.columns.  Both are non of non-trivial size.  And both are nearly-repeated in multiple files.
Ideally, we should find ways to extract the logic for reuse.

But for now, can you simply move the printing code to a separate function here to split the responsibilities?

cmd/ocm/list/org/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+			// Map used to parse Organization data -
+			// Writer to body variable:
+			var body string
+			var jsonBody map[string]interface{}
+			boddyBuffer := bytes.NewBufferString(body)
```
cben, line 118: s/boddy/body/g

cmd/ocm/list/org/cmd.go
cben: [out of scope, just observation: it's funny we're using the SDK "typed" interface to deserialize response into structs, only to ser/deserialize again into interface{}.  I suspect it'd be _easier_ to extract generic page-walking and column-printing logics if we were using untyped `connect.Get()`.  
Maybe the best reuse unit could be a "half-typed" interface that did deserialize top-level `.Total`, `.Page`, etc but left `Items []interface{}`, which would allow reusing _a common result type for all collections_ (together with `Error` type which is already common).]

cben:
Other than small nits, looks good, and works too :+1:

cmd/ocm/list/org/cmd.go
cben: [continuing my braindump: SDK could also use `json.RawMessage` to _delay the decision_ whether to unmarshal Items as typed or map[string]interface{} :gift:  https://eli.thegreenplace.net/2020/representing-json-structures-in-go/#hybrid-representation]

cben:
I see this got stuck.  Let's merge this as-is :+1:, I can do a follow-up if I care.

PR #180, opened 2020-11-15
ocm create cluster: Honor --parameter and --header flags

cben:
In particular, `--parameter=dryRun=true` now works.

In case of no errors, results in 204 No Content, in which case I added a check to print nothing (rather than descrption full of zero values):
```
$ go run ./cmd/ocm create cluster beni-6 --parameter=dryRun=true
$ go run ./cmd/ocm create cluster beni-6 --parameter=dryRun=true --region=wrong
Error: Failed to create cluster: unable to create cluster: identifier is '400', code is 'CLUSTERS-MGMT-400' and operation identifier is '1gvgbb2a0ao9iuo75ps6mbq6fc961ktn': Region 'wrong' is not supported for cloud provider 'aws'
exit status 1
```

in `--debug` mode this still confuses the SDK — opened https://github.com/openshift-online/ocm-sdk-go/issues/296:
```
I1115 18:02:10.756710  792371 dump.go:143] Response status is '204 No Content'
I1115 18:02:10.756771  792371 dump.go:155] Response header 'Content-Encoding' is 'gzip'
I1115 18:02:10.756801  792371 dump.go:155] Response header 'Content-Type' is 'application/json'
I1115 18:02:10.756854  792371 dump.go:155] Response header 'Date' is 'Sun, 15 Nov 2020 16:02:10 GMT'
I1115 18:02:10.756882  792371 dump.go:155] Response header 'Vary' is 'Accept-Encoding'
I1115 18:02:10.756908  792371 dump.go:155] Response header 'X-Operation-Id' is '1gvg60n0dol7pmtv0r60hiftiqjdkt86'
I1115 18:02:10.756935  792371 dump.go:159] Response body follows
I1115 18:02:10.756974  792371 dump.go:249] 
Error: Failed to create cluster: unable to create cluster: ReadObject: expect { or , or } or n, but found , error found in #0 byte of ...||..., bigger context ...||...
exit status 1
```

But still much better than ignoring the param and creating a cluster :money_with_wings:...

### DIFF
--- a/cmd/ocm/cluster/create/create.go
+++ b/cmd/ocm/cluster/create/create.go
@@ -226,9 +226,11 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 
 	// Send a request to create the cluster:
-	response, err := cmv1Client.Clusters().Add().
-		Body(cluster).
-		Send()
+	request := cmv1Client.Clusters().Add().
+		Body(cluster)
+	arguments.ApplyParameterFlag(request, args.parameter)
+	arguments.ApplyHeaderFlag(request, args.header)
+	response, err := request.Send()
 	if err != nil {
 		return fmt.Errorf("unable to create cluster: %v", err)
 	}
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -307,7 +307,7 @@ func run(cmd *cobra.Command, argv []string) error {
 		Private:            &args.private,
 	}
 
-	cluster, err := c.CreateCluster(cmv1Client, clusterConfig)
+	cluster, err := c.CreateCluster(cmv1Client, clusterConfig, args.parameter, args.header)
 	if err != nil {
 		return fmt.Errorf("Failed to create cluster: %v", err)
 	}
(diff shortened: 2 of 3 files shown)


cmd/ocm/cluster/create/create.go
```diff
@@ -226,9 +226,11 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 
 	// Send a request to create the cluster:
-	response, err := cmv1Client.Clusters().Add().
-		Body(cluster).
-		Send()
+	request := cmv1Client.Clusters().Add().
+		Body(cluster)
+	arguments.ApplyParameterFlag(request, args.parameter)
```
igoihman, line 231: I prefer we dryRun as command-line parameter as I am trying to separate the API-like commands from other commands. Same as in https://github.com/openshift/moactl/blob/master/cmd/create/cluster/cmd.go#L51 for reference.

igoihman:
Thanks for this change!

cmd/ocm/cluster/create/create.go
cben: I thought about that too, sounds good.  Should I remove the non-working --parameter & --header from this command then?

igoihman: yes, good idea

cben:
Closing in favor of #182.

PR #182, opened 2020-11-16
ocm create cluster --dry-run

cben:
Alternative to #180.

1. Drops --parameter and --header flags from `cluster create` and `create cluster` commands.
   These never worked in these 2 commands, and it makes sense to only have them in lower-level REST commands like `ocm post`.

2. Adds `ocm create cluster --dry-run` flag.

```
$ go run ./cmd/ocm create cluster beni-7 --dry-run --region=wrong
Error: Failed to create cluster: dry run: unable to create cluster: identifier is '400', code is 'CLUSTERS-MGMT-400' and operation identifier is '1h007pdkir8834u880ea7bblut6nan3k': Region 'wrong' is not supported for cloud provider 'aws'
exit status 1

$ go run ./cmd/ocm create cluster beni-7 --dry-run 
dry run: Would be successful.
```

In debug mode, the successful case shows a confusing `Error: Failed to create cluster: dry run: unable to create cluster: ReadObject: expect { or , or } or n, but found , error found in #0 byte of ...||..., bigger context ...||...` error, but it's still dry, no cluster gets created, and that error will be fixed by https://github.com/openshift-online/ocm-sdk-go/pull/297

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -26,14 +26,11 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/util/sets"
 
-	"github.com/openshift-online/ocm-cli/pkg/arguments"
 	c "github.com/openshift-online/ocm-cli/pkg/cluster"
 	"github.com/openshift-online/ocm-cli/pkg/ocm"
 )
 
 var args struct {
-	parameter          []string
-	header             []string
 	region             string
 	version            string
 	flavour            string
@@ -70,8 +67,6 @@ var Cmd = &cobra.Command{
 
 func init() {
 	fs := Cmd.Flags()
-	arguments.AddParameterFlag(fs, &args.parameter)
-	arguments.AddHeaderFlag(fs, &args.header)
 	fs.StringVar(
 		&args.region,
 		"region",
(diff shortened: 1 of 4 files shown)


cmd/ocm/create/cluster/cmd.go
```diff
@@ -26,14 +26,13 @@ import (
...
-	"github.com/openshift-online/ocm-cli/pkg/arguments"
 	c "github.com/openshift-online/ocm-cli/pkg/cluster"
 	"github.com/openshift-online/ocm-cli/pkg/ocm"
 )
 
 var args struct {
-	parameter          []string
-	header             []string
+	dryRun bool
+
```
igoihman, line 35: redundant line

cben: It's a "how to create" flag, I deliberately left empty lines to separate it from "what cluster to create".

cben:
```
cmd/ocm/create/cluster/cmd.go:33:10: struct of size 328 bytes could be of size 320 bytes (maligned)
```
Any objection to turning off `maligned` linter?  I think we can afford 8 bytes of RAM and not worry about such micro-optimizations in this project.

igoihman:
@cben no objection

cben:
Green, PTAL.

PR #183, opened 2020-11-16
remove region display name

igoihman:
We are planning to remove the display name from the UI so I'd like the CLI to match the same behavior


igoihman:
@cben PTAL

cben:
Oh well.  IMHO it was useful but one can always get this info from Amazon / Google. And it's still in the json...

Code LGTM 👍

PR #187, opened 2020-11-23
Use Must() wrapper to panic in case of wrong flag names in code

cben:
@igoihman @vkareh please review.

gosec has a point, this is better than just discarding the error.  Now in case of a typo in flag name in code, we'll get something like:
```
$ go run ./cmd/ocm whatever
panic: flag "xpiration-time" does not exist

goroutine 1 [running]:
github.com/openshift-online/ocm-cli/pkg/arguments.Must(...)
	/home/cben/openshift/ocm-cli/pkg/arguments/arguments.go:40
github.com/openshift-online/ocm-cli/cmd/ocm/edit/cluster.init.0()
	/home/cben/openshift/ocm-cli/cmd/ocm/edit/cluster/cmd.go:77 +0x2b4
exit status 2
```

Future TODO: This is not yet caught by `make test` and would not fail CI, but it wouldn't before this PR either.
It'll become covered if we add docs or completions generation like in moactl.

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -111,16 +111,14 @@ func init() {
 		args.expirationTime,
 		"Specified time when cluster should expire (RFC3339). Only one of expiration-time / expiration may be used.",
 	)
-	//nolint:gosec
-	fs.MarkHidden("expiration-time")
+	arguments.Must(Cmd.Flags().MarkHidden("expiration-time"))
 	fs.DurationVar(
 		&args.expirationSeconds,
 		"expiration",
 		args.expirationSeconds,
 		"Expire cluster after a relative duration like 2h, 8h, 72h. Only one of expiration-time / expiration may be used.",
 	)
-	//nolint:gosec
-	fs.MarkHidden("expiration")
+	arguments.Must(Cmd.Flags().MarkHidden("expiration"))
 	fs.BoolVar(
 		&args.private,
 		"private",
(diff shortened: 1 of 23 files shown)


cben:
@igoihman ping, please review.

cben:
@vkareh please review.

vkareh:
This is a good idea. I wonder if we can provide some utility functions to make it easier on the developer. Currently, any reference to a flag, for example
```go
if cmd.Flags().Changed("private") {...}
```
would then need to be wrapped. A wrapper for `cmd.Flags()` would solve the issue and future proof the code. What do you think?

cben:
@vkareh The way I see it, `Must()` is such a utility function :wink:  
I'm not sure what you mean by a wrapper for `cmd.Flags()`.  That returns a `pflag.FlagSet`, which has 2 metric tons of methods :-(  

* We could wrap only `FlagSet.Lookup()` with a helper that returns a `Flag` or panics, then change some `Cmd` and the `FlagSet` calls to use the Flag instead.  However not all functionality is available/equally convenient via the Flag...

* specifically `cmd.Flags().Changed()` doesn't need much wrapping, has "If a flag doesn't exist, it wasn't changed...." behavior which is good enough IMO.
https://github.com/spf13/pflag/blob/85dd5c8bc61cfa382fecd072378089d4e856579d/flag.go#L517-L520

TBH part of the reason I did this was specifically the gosec pragmas ;-)  
Any objection to merging this?

PR #190, opened 2020-11-25
Add a url option to tunnel command

wanghaoran1988:
This add url option to tunnel command, so we can override the existing api gateway url in the ocm.json.

### DIFF
--- a/cmd/ocm/tunnel/cmd.go
+++ b/cmd/ocm/tunnel/cmd.go
@@ -24,10 +24,15 @@ import (
 	"strings"
 
 	c "github.com/openshift-online/ocm-cli/pkg/cluster"
+	"github.com/openshift-online/ocm-cli/pkg/config"
 	"github.com/openshift-online/ocm-cli/pkg/ocm"
 	clustersmgmtv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 )
+
+var args struct {
+	url string
+}
 
 var Cmd = &cobra.Command{
 	Use:   "tunnel [flags] {CLUSTERID|CLUSTER_NAME|CLUSTER_NAME_SEARCH} -- [sshuttle arguments]",
@@ -42,6 +47,13 @@ var Cmd = &cobra.Command{
 }
 
 func init() {
+	flags := Cmd.Flags()
+	flags.StringVar(
+		&args.url,
+		"url",
+		"",
+		"URL of the API gateway.",
+	)
 }
 
 func run(cmd *cobra.Command, argv []string) error {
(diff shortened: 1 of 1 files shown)


cmd/ocm/tunnel/cmd.go
```diff
@@ -65,14 +77,21 @@ func run(cmd *cobra.Command, argv []string) error {
...
-
 	path, err := exec.LookPath("sshuttle")
 	if err != nil {
 		return fmt.Errorf("to run this, you need install the sshuttle tool first")
 	}
 
 	// Create the client for the OCM API:
-	connection, err := ocm.NewConnection().Build()
+	config, err := config.Load()
+	if err != nil {
```
feichashao, line 87: Also need to check if config == nil, as config.Load() might return an empty object.
```
 ./ocm tunnel hive-stage-01 --url https://api.openshift.com 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x88 pc=0x9d7c3a]

goroutine 1 [running]:
github.com/openshift-online/ocm-cli/cmd/ocm/tunnel.run(0x1154120, 0xc0001eb8f0, 0x1, 0x3, 0x0, 0x0)
	/home/siwu/git/haoran/ocm-cli/cmd/ocm/tunnel/cmd.go:91 +0x19a
github.com/spf13/cobra.(*Command).execute(0x1154120, 0xc0001eb8c0, 0x3, 0x3, 0x1154120, 0xc0001eb8c0)
	/home/siwu/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842 +0x453
github.com/spf13/cobra.(*Command).ExecuteC(0x1154900, 0xc000108060, 0x4, 0x4)
	/home/siwu/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x349
github.com/spf13/cobra.(*Command).Execute(...)
	/home/siwu/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main()
	/home/siwu/git/haoran/ocm-cli/cmd/ocm/main.go:129 +0x1ec

$ ls ~/.ocm.json 
ls: cannot access '/home/siwu/.ocm.json': No such file or directory
```

wanghaoran1988: done

feichashao: works!

wanghaoran1988:
cc @cblecker

jhernand:
@wanghaoran1988 Can you explain why is this option needed? Please note that if you just override the OCM URL and not the other parameters (in particular existing tokens) then you may be connecting to a server with the credentials for other server. That isn't currently a real problem because we use the same SSO servers for almost all environments (except for the SSO test environment). But this can change at any time; we may want to change the stage environment, for example, to use a different SSO service. If we do so and you use this `--url` mechanism you could be connecting to the stage environment inadvertently using the credentials of the production environment.

If what you want to do is to switch from one environment to another without having to repeat the `ocm login ...` command each time you can use the `OCM_CONFIG` environment variable as described [here](https://github.com/openshift-online/ocm-cli#multiple-concurrent-logins-with-ocm_config).

PR #192, opened 2020-11-26
GCP CCS command

igoihman:


cmd/ocm/create/cluster/cmd.go
```diff
@@ -75,6 +80,12 @@ func init() {
 
 	arguments.AddProviderFlag(fs, &args.provider)
 	arguments.AddCCSFlags(fs, &args.ccs)
+
+	fs.Var(
+		&args.gcpServiceAccount,
```
cben, line 85: Ideally declaring this flag would be part of `arguments.AddCCSFlags()`.  
One way to do that that might be move the FilePath var inside CCS struct, and making `constructGCPCredentials` a method on CCS struct.  Don't know if worth it.

cben:
Looks great :+1:

cmd/ocm/create/cluster/cmd.go
cben: Opened #199 for some validations we should improve.

cmd/ocm/create/cluster/cmd.go
```diff
@@ -75,6 +80,12 @@ func init() {
...
+		"service-account-file",
```
cben, line 86: WDYT about renaming this `--gcp-service-account-file`?  4 more characters to type but maybe worth it to keep the flags organized.

(BTW, maybe it's time to start using Cobra's flag groups to organize the help :thinking:)

pkg/arguments/arguments.go
```diff
@@ -26,14 +26,33 @@ import (
 	"reflect"
 	"strings"
 
-	"github.com/mattn/go-isatty"
+	isatty "github.com/onsi/ginkgo/reporters/stenographer/support/go-isatty"
```
cben, line 29: @igoihman was this deliberate?
https://github.com/onsi/ginkgo/tree/master/reporters/stenographer/support/go-isatty/
looks like an old, less maintained, copy of upstream https://github.com/mattn/go-isatty/.
Or was this simply replaced by IDE being too clever?

cben: => #201 if accidental.

PR #195, opened 2020-11-29
Validate SSH Config file for ProxyJump

jharrington22:
For `ocm tunnel` validate that the user has their SSH config setup to proxy `devshift.org` and `openshiftapps.com` domains to another host, eg. bastion.

Its a pretty simple validation as it only checks that your SSH config file contains these two domains it doesn't validate that the SSH config block is ProxyJump.

This is to help notify the user that their SSH config isn't setup correct and prevents long timeouts when you don't have it setup since you cannot SSH to `rh-api.*(openshiftapps.com|devshift.org)` domains without proxying through a bastion host.

### DIFF
--- a/cmd/ocm/tunnel/cmd.go
+++ b/cmd/ocm/tunnel/cmd.go
@@ -17,9 +17,12 @@ limitations under the License.
 package tunnel
 
 import (
+	"bufio"
 	"fmt"
+	"log"
 	"os"
 	"os/exec"
+	"os/user"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


cben:
Can you get somebody else from SRE to review?  This command is effectively maintained by SREs anyway, and I think PRs will land faster and get better feedback if you cross-review...

Some thoughts from me, possibly silly:

- On errors, consider printing an example of what the config _should_ look like.
- Is the needed config fixed?  Could you use `ssh -F configfile` to make it Just Work instead of requiring user to config? :thinking: 
- Could you avoid parsing ssh config and instead _probe_ whether it's good by running some `ssh` commands?

But if you believe you need this logic (and want to maintain it :stuck_out_tongue:), looks reasonable.

PR #196, opened 2020-11-30
get service account file only for GCP CCS

igoihman:


cmd/ocm/create/cluster/cmd.go
```diff
@@ -333,16 +333,19 @@ func fetchFlavours(client *cmv1.Client) (flavours []*cmv1.Flavour, err error) {
...
-	jsonFile, err := os.Open(filePath.String())
-	if err != nil {
-		return err
-	}
-	defer jsonFile.Close()
-	byteValue, _ := ioutil.ReadAll(jsonFile)
-	err = json.Unmarshal(byteValue, &value.GCP)
-	if err != nil {
-		return err
+	if value.Enabled {
```
cben, line 336: This func is only called when the flag is set (`!= ""`).
This change makes it easier to _ignore_ this flag when not relevant, right?
I propose instead returning error when it's set but not relevant — see #199.  WDYT?

PR #205, opened 2020-12-30
Dynamic completion of flag values --region= etc.

cben:
Improved shell completion — can now complete some flag values based on results from server --provider, --region, --version, --flavour, --compute-machine-type.
In some shells (fish, zsh) can show extra info for the completions :sparkles:.  In --interactive mode the descriptions are not used (not currently supported by `survey` library).
[asciicast](https://asciinema.org/a/r6ZSapnhpBP0k8CGR2alDwwso)

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -25,10 +25,10 @@ import (
 	"time"
 
 	"github.com/AlecAivazis/survey/v2"
+	sdk "github.com/openshift-online/ocm-sdk-go"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/openshift-online/ocm-cli/pkg/arguments"
 	c "github.com/openshift-online/ocm-cli/pkg/cluster"
@@ -89,6 +89,7 @@ func init() {
 	)
 
 	arguments.AddProviderFlag(fs, &args.provider)
+	Cmd.RegisterFlagCompletionFunc("provider", arguments.MakeCompleteFunc(osdProviderOptions))
 	arguments.AddCCSFlags(fs, &args.ccs)
 
 	fs.Var(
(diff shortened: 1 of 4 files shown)


cben:
@igoihman @jhernand @nimrodshn please review.

cben:
Added --compute-machine-types completion that I'd missed.  Rebased.

cben:
@nimrodshn could you please test this too? (and/or review)

@igoihman started testing but only old completion of ocm commands & --flag-names worked.  For me it does work after doing what `ocm completion --help` says, and we're not sure what's going on :confused:.

cmd/ocm/create/cluster/cmd.go
```diff
@@ -196,8 +206,72 @@ func init() {
...
+	options := make([]arguments.Option, len(regions))
+	for i, region := range regions {
+		// `enabled` flag only affects Red Hat infra. All regions enabled on CCS.
+		if args.ccs.Enabled || region.Enabled() {
+			options[i] = arguments.Option{
+				Value:       region.ID(),
+				Description: region.DisplayName(),
+			}
+		}
+	}
```
nimrodshn, line 230: @cben Wouldn't this pre-allocate a potentially larger "options" slice? if not all the regions are available as options? Perhaps `survey` knows how to deal with this but it just caught my eye.

nimrodshn:
@cben please add some docs on using the dynamic completion `ocm completion` etc.

Tested locally -  works well :+1: 

Aprart from ^^ comments LGTM :100:

cmd/ocm/create/cluster/cmd.go
cben: Oops :flushed: great catch!  Fixed.

cben:
@igoihman We figured with Nimrod that activating the completions by `go install` is error-prone if you have another `ocm` in PATH, e.g. from rpm.
A more robust recipe for testing is running it from current dir:
```
export PATH="$PWD:$PATH"
make && source <(ocm completion bash)
which -a ocm  # just to make sure
```
but yes, I definitely need to revisit the help...

igoihman:
Thanks @cben and @nimrodshn . I tested the change and it works great!

igoihman:
merging as is - @cben please add more info to the help in another PR

PR #207, opened 2021-01-03
print default version in version list

igoihman:
```
$ ocm list versions --default
4.6.9


$ ocm list versions -h
List the versions available for provisioning a cluster

Usage:
  ocm list versions [flags]

Aliases:
  versions, version

Examples:
  # List all supported cluster versions
  ocm list versions

Flags:
  -d, --default   Show the default version
  -h, --help      help for versions

```

### DIFF
--- a/cmd/ocm/list/version/cmd.go
+++ b/cmd/ocm/list/version/cmd.go
@@ -24,6 +24,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var args struct {
+	defaultVersion bool
+}
+
 var Cmd = &cobra.Command{
 	Use:     "versions",
 	Aliases: []string{"version"},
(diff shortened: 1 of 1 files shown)


nimrodshn:
LGTM :+1:

cmd/ocm/list/version/cmd.go
```diff
@@ -35,6 +39,17 @@ var Cmd = &cobra.Command{
...
 }
 
+func init() {
+	fs := Cmd.Flags()
+	fs.BoolVarP(
+		&args.defaultVersion,
+		"default",
+		"d",
+		false,
+		"Show the default version",
```
cben, line 49: Consider something like "Show the default version instead of listing all versions".

cben: or "Show only the default version".

cben: LGTM otherwise +1

PR #213, opened 2021-01-14
Avoid `survey.Select` bug when Default is not one of Options

cben:
Before, this was possible if user chooses gcp then presses Enter immediately on "Region:" propmt (without moving cursor):
```
$ ./ocm create cluster foo --dry-run --interactive
? Cloud provider: gcp
? CCS: No
? Multiple AZ: No
? Region: us-east-1
Error: if specifying a non-aws cloud provider, region must be set to a valid region
```
(despite "us-east-1" NOT being one of the gcp options!  known upstream issue https://github.com/AlecAivazis/survey/pull/284)

Now with gcp the region prompt gets no default so Enter selects the first option ("asia-east1").

Partially addresses https://issues.redhat.com/browse/SDA-3524

### DIFF
--- a/pkg/arguments/interactive.go
+++ b/pkg/arguments/interactive.go
@@ -29,6 +29,7 @@ import (
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 const questionAnnotationKey = "flag_survey_question"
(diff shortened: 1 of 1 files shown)


cben:
@igoihman @nimrodshn @vkareh please review.

I still need to decide whether to make a per-provider default (`us-east-1` vs `us-east1`) as https://issues.redhat.com/browse/SDA-3524 suggests, or just drop the default and require user to always decide a region.
But anyway this makes the interactive.go helper more robust.

igoihman:
@cben I'm OK with dropping the default and let the user pick the region

igoihman:
LGTM

PR #215, opened 2021-01-17
Drop default of --region

cben:
It was only appropriate for AWS and it's good for users to make a conscious decision before creating clusters.

(This is the last part of addressing https://issues.redhat.com/browse/SDA-3524)

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -75,7 +75,8 @@ var Cmd = &cobra.Command{
 	Long: fmt.Sprintf("Create managed OpenShift Dedicated v4 clusters via OCM.\n"+
 		"\n"+
 		"NAME %s", clusterNameHelp),
-	RunE: run,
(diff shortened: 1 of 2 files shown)


cben:
2nd commit made some changes to allow the standard `Cmd.MarkFlagRequired("region")`.
Now prints `Error: required flag(s) "region" not set`.

- Moved flag validation & interactive prompts from `RunE` to `PreRunE`.
  This allows interactive helpers to .Set flags that user didn't pass.
  (Cobra enforces required flags between PreRunE & RunE.)

- Fixed interactive helpers to use higher-level `FlagSet.Set()` that
  also marks them `.Changed`.

cben:
@igoihman @nimrodshn please review

cben:
@nimrodshn @vkareh @igoihman @jhernand please review.

vkareh:
Yes, this makes sense

/lgtm

cben:
ping can this be merged?

PR #217, opened 2021-01-27
Add autoscaling to ocm cli

zgalor:
Commands updated:
ocm list machinepools - added autoscaling field, and range
ocm create cluster - added autoscaling params
ocm describe cluster - added autoscaling indication and range
ocm edit machinepool - can now edit default machine pool - and autoscaling params
ocm edit cluster - no longer able to edit compute nodes


zgalor:
@igoihman @cben @vkareh @nimrodshn @jhernand PTAL

cmd/ocm/edit/machinepool/cmd.go
```diff
@@ -146,7 +173,52 @@ func run(cmd *cobra.Command, argv []string) error {
...
+			return fmt.Errorf("--replicas can't be set with autoscaling paramteres")
+		}
+	}
+
+	// Editing the default machine pool is a different process
+	if machinePoolID == "default" {
+		clusterConfig := c.Spec{
+			Autoscaling: c.Autoscaling{
+				Enabled:     args.autoscalingEnabled,
+				MinReplicas: args.minReplicas,
```
igoihman, line 195: please add validation checks for min/max replicas values

zgalor: ocm cli edit didn't validate any of the edit values before this change. It relies on clusters service validations.

Currently ocm cli only validates the values in cluster create.

That is because validation is more complex and requires getting the cluster and the machine pool.

Do we want to validate these values in ocm-cli?

zgalor: @igoihman I added validations for the default machine pool.

Before the validations this is the error the user got:
```
Error: Failed to update machine pool 'default' on cluster 'zg-j31': identifier is '400', code is 'CLUSTERS-MGMT-400' and operation identifier is '1iim11mj6epg0lu178m4tin487rhu91d': Invalid number of compute nodes: 1. Minimal requirement is 2 compute nodes.
```

This is the error after the validation:
```
Error: Minimum is 2 nodes

```

I am not sure if the validation is required.
It might even be confusing for the user that only wants to update max-replicas on a machine pool that has autoscaling 4-6 nodes.
If the user tries to set max-replicas=1 they will get an error that the minimum is 2 (but the minimum in this case would be 4 - the num of min-replicas).

Since it's a complex validation to check what changed, in CS we first update the DB - validate the latest values, and if an error occurs we rollback - which obviously we can't do in the CLI.

I didn't validate for the additional machine pool - which is even more complex

Thinking again I think I should remove the max-replicas validation I just added. WDYT?

cmd/ocm/cluster/create/create.go
```diff
@@ -231,7 +231,7 @@ func run(cmd *cobra.Command, argv []string) error {
 	// Print the result:
 	cluster = response.Body()
 
-	err = clusterpkg.PrintClusterDesctipion(connection, cluster)
+	err = clusterpkg.PrintClusterDescripion(connection, cluster)
```
nimrodshn, line 234: **PrintClusterDescription

zgalor: Fixed. Thanks

pkg/cluster/describe.go
```diff
@@ -103,6 +103,16 @@ func PrintClusterDesctipion(connection *sdk.Connection, cluster *cmv1.Cluster) e
 		shard = shardPath.Body().HiveConfig().Server()
 	}
 
+	var computesStr string
+	if cluster.Nodes().AutoscaleCompute() != nil {
+		computesStr = fmt.Sprintf("%d-%d (Autoscaled)",
```
nimrodshn, line 108: Wonder if theres a better way to present this; `Min: x, Max: y`.

Not entirely sure

nimrodshn:
LGTM :100:

pkg/cluster/describe.go
zgalor: I wrote like this because it aligns with `rosa describe cluster` and `list macinepools`

cmd/ocm/create/machinepool/cmd.go
```diff
@@ -40,6 +43,9 @@ var Cmd = &cobra.Command{
 	Long:    "Add a machine pool to the cluster.",
 	Example: `  # Add a machine pool mp-1 with 3 replicas and m5.xlarge instance type to a cluster
   ocm create machinepool --cluster mycluster --instance-type m5.xlarge --replicas 3 mp-1
+  # Add a machine pool mp-1 with autoscaling enabled and 3 to 6 replicas of m5.xlarge to a cluster
+  ocm create machinepool --cluster=mycluster --name=mp-1 --enable-autoscaling
```
vkareh, line 47: You should end this line with `\`, otherwise a user might copy-paste and end up running two separate commands

cmd/ocm/edit/machinepool/cmd.go
```diff
@@ -61,6 +66,27 @@ func init() {
...
+		0,
+		"Minimum number of machines for the machine pool.",
+	)
+
+	flags.IntVar(
+		&args.maxReplicas,
+		"max-replicas",
+		0,
+		"Maximum number of machines for the machine pool.",
+	)
```
vkareh, line 88: Can you use `arguments.AddAutoscalingFlags(...)` here?

cmd/ocm/create/machinepool/cmd.go
zgalor: Fixed

cmd/ocm/edit/machinepool/cmd.go
zgalor: I can.
That would require changing the implementation to be like the cluster - Autoscaling struct.
I should probably do the same for `create`

vkareh:
LGTM

PR #220, opened 2021-02-09
Fix create cluster command with autoscaling

zgalor:
There should be defaults for min/max replicas only for interactive mode.
Otherwise, they are required when --enable-autoscaling is set.

Related: https://issues.redhat.com/browse/SDA-3702

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -385,7 +385,7 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		return err
 	}
 
-	err = arguments.CheckIgnoredAutoscalingFlags(args.autoscaling, args.computeNodes)
+	err = arguments.CheckAutoscalingFlags(args.autoscaling, args.computeNodes)
 	if err != nil {
 		return err
 	}
@@ -622,15 +622,19 @@ func promptAutoscaling(fs *pflag.FlagSet) error {
 		return err
 	}
 	if args.autoscaling.Enabled {
-		// set default
-		args.autoscaling.MinReplicas = minComputeNodes(args.ccs.Enabled, args.multiAZ)
+		// set default for interactive mode
+		if args.interactive && args.autoscaling.MinReplicas == 0 {
+			args.autoscaling.MinReplicas = minComputeNodes(args.ccs.Enabled, args.multiAZ)
+		}
 		err = arguments.PromptInt(fs, "min-replicas", validateAutoscalingMin)
 		if err != nil {
 			return err
 		}
 
-		// set default
-		args.autoscaling.MaxReplicas = args.autoscaling.MinReplicas
+		// set default for interactive mode
+		if args.interactive && args.autoscaling.MaxReplicas == 0 {
+			args.autoscaling.MaxReplicas = args.autoscaling.MinReplicas
+		}
 		err = arguments.PromptInt(fs, "max-replicas", validateAutoscalingMax)
 		if err != nil {
 			return err
--- a/pkg/arguments/arguments.go
+++ b/pkg/arguments/arguments.go
@@ -183,9 +183,11 @@ func AddAutoscalingFlags(fs *pflag.FlagSet, value *cluster.Autoscaling) {
 	SetQuestion(fs, "max-replicas", "Max replicas:")
 }
 
-// CheckIgnoredAutoscalingFlags errors if --min-replicas or --max-replicas
+// CheckAutoscalingFlags errors if --min-replicas or --max-replicas
 // were used without --enable-autoscaling (and vice-versa with --compute-nodes)
-func CheckIgnoredAutoscalingFlags(autoscaling cluster.Autoscaling, computeNodes int) error {
+// It also errors if --min-replicas or --max-replicas were not supplied
+// when --enable-autoscaling is used
+func CheckAutoscalingFlags(autoscaling cluster.Autoscaling, computeNodes int) error {
 	if !autoscaling.Enabled {
 		bad := []string{}
 		if autoscaling.MinReplicas != 0 {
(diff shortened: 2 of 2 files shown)


zgalor:
@cben @nimrodshn @vkareh @jhernand @igoihman PTAL

nimrodshn:
LGTM :100:

cmd/ocm/create/cluster/cmd.go
```diff
@@ -622,15 +622,19 @@ func promptAutoscaling(fs *pflag.FlagSet) error {
 		return err
 	}
 	if args.autoscaling.Enabled {
-		// set default
-		args.autoscaling.MinReplicas = minComputeNodes(args.ccs.Enabled, args.multiAZ)
+		// set default for interactive mode
+		if args.interactive {
```
cben, line 626: Is the problem that the flag should have no default at all when not interactive, or that you were overriding the flag value even when passed from command-line?

Some tips about implementation: (a bit subtle, so want to confirm desired behavior ^^ before I comment on implementation)

- `fs.Changed("min-replicas")` can tell you whether flag was passed from command line.
- _when not Changed()_, writing directly to a flag variable `args.autoscaling.MinReplicas = ...` effectively sets the default, which is useful for interactive prompts.
- when Changed(), interactive code should not touch the value.
- `fs.Set("min-replicas", "13")` would both change the variable and mark it Changed, so prompt will be skipped.  Don't want that.

pkg/arguments/arguments.go
```diff
@@ -204,6 +204,14 @@ func CheckIgnoredAutoscalingFlags(autoscaling cluster.Autoscaling, computeNodes
 		if computeNodes != 0 {
 			return fmt.Errorf("--compute-nodes is meaningless with --enable-autoscaling")
 		}
+
+		if autoscaling.MinReplicas == 0 {
+			return fmt.Errorf("--min-replicas flag is required with --enable-autoscaling")
+		}
+
+		if autoscaling.MaxReplicas == 0 {
+			return fmt.Errorf("--max-replicas flag is required with --enable-autoscaling")
```
cben, line 213: consider renaming `CheckIgnoredAutoscalingFlags` to `CheckAutoscalingFlags`.
The "CheckIgnored..." functions originally were about flag combinations where otherwise one of them would be ignored.

cmd/ocm/create/cluster/cmd.go
zgalor: So it's both.
It not interactive I want to force the user to provide it.
And If it's interactive I want to cgane the default (that is 0 when non set) to the minimum allowed.

pkg/arguments/arguments.go
zgalor: Agree

zgalor: fixed

cmd/ocm/create/cluster/cmd.go
cben: I think this code still overrides command line values when --interactive is given :-1:
It is a feature that you can provide any subset of flags with --interactive, and these will be used without prompting.

One simple fix is to change the condition to `if args.autoscaling.MinReplicas == 0` or `if !fs.Changed("min-replicas") {`.
**That would also apply the minComputeNodes and MinReplicas defaults when not interactive.**  
If you go this way, you can also drop required checks from CheckIgnoredAutoscalingFlags.

cben: Ah, only now saw your reply.
> If not interactive I want to force the user to provide it.

In that case, I think you want `if args.interactive && args.autoscaling.MinReplicas == 0 {`.

cmd/ocm/create/cluster/cmd.go
```diff
@@ -385,7 +385,7 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		return err
 	}
 
-	err = arguments.CheckIgnoredAutoscalingFlags(args.autoscaling, args.computeNodes)
+	err = arguments.CheckAutoscalingFlags(args.autoscaling, args.computeNodes)
```
cben, line 388: It's confusing that this is called after `promptAutoscaling` but before prompting for `computeNodes`, yet checks both of them...
But that's out of scope for this PR, I'll send a followup.

cmd/ocm/create/cluster/cmd.go
zgalor: fixed

pkg/arguments/arguments.go
```diff
@@ -204,6 +206,14 @@ func CheckIgnoredAutoscalingFlags(autoscaling cluster.Autoscaling, computeNodes
 		if computeNodes != 0 {
 			return fmt.Errorf("--compute-nodes is meaningless with --enable-autoscaling")
 		}
+
+		if autoscaling.MinReplicas == 0 {
```
vkareh, line 210: Do we want to check `<= 0` just so that we can catch those before the API does?

vkareh (commented):
Other than a question/comment about validation, this LGTM

pkg/arguments/arguments.go
cben: If we do, it's better to put in a validation callback, like `validateComputeNodes`, so that interactively it re-prompts if typed out of range value.  Although there is presently an annoying bug with re-prompting (prev text remains there but not visible).

Anyway, out of scope for this PR.

cben: BTW, how does auto-scaling work with multi-AZ?  Are the min/max per-AZ or * 3 for total?  Does backend reject non-multiples of 3?

(--compute-nodes currently takes total (multiple of 3), while UI takes per-AZ number.)

cben:
LGTM :100:

pkg/arguments/arguments.go
zgalor: @vkareh we validate min (which can be 2,3,4 or 9 according to multi-az and ccs booleans).

@cben autoscaling works the same with multi-az as compute nodes. Both min and max must be a multiple of 3.

Theres a validation callback for min that checks min number and multi az - `validateAutoscalingMin`
and a validator for max checking its >= min and multi az constraint - `validateAutoscalingMin`
Both sent to `promptInt`

PR #222, opened 2021-02-09
fix cluster admin enabled output

igoihman:


pkg/cluster/describe.go
```diff
@@ -113,6 +113,18 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
 		computesStr = fmt.Sprintf("%d", cluster.Nodes().Compute())
 	}
 
+	clusterAdminEnabled := false
+	if cluster.CCS().Enabled() {
+		clusterAdminEnabled = true
```
cben, line 118: Would be nice if the backend had a single place you can check for "do I have this ability".
e.g. add this label for all CCS clusters, or a top-level boolean field.  Or maybe an authorization request.

BTW, does this mean user _has_ admin, or _can enable_ admin?  The "Cluster Admin:" label is not very clear.

(I'm not sure how exactly admin works)

cben: Not objection to adding this code here, just pointing out it's an "API smell"

zgalor: It means the user can see the `cluster-admins` group, add and remove users from it, and it will be bound to admin role on the cluster

pkg/cluster/describe.go
```diff
@@ -113,6 +113,18 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
 		computesStr = fmt.Sprintf("%d", cluster.Nodes().Compute())
 	}
 
+	clusterAdminEnabled := false
```
zgalor, line 116: This entire code should go inside `func isClusterAdminEnabled`.
We will also need it for user add/delete commands.

zgalor: Well. Not that critical. I see you will need to pass the subscription response.
We can refactor later when it's actually used.

pkg/cluster/describe.go
cben: Then maybe reword it "Cluster admin possible:" or "Cluster admin allowed:" or something like that.

What if `describe cluster` just tries to get `groups/cluster-admins` instead of making policy assumptions?

cben: Does this info belong in `describe cluster`?
Is it visible in `ocm list users`, which can also answer which users were granted admin?

here is a CCS cluster with some cluster admins:
```
$ ocm  --cluster=1ave4huo9pdk0qqt6tqu92mktivs5c0j list users
GROUP               USER
dedicated-admins    OpenShift Dedicated
dedicated-admins    ahoyun
dedicated-admins    sysposddadmin
cluster-admins    sysposdcadmin
```
here is different cluster where the group exists but is empty:
```
{
  "kind": "Group",
  "href": "/api/clusters_mgmt/v1/clusters/15uq9gvgj178aumrphhba0nr24j75md0/groups/cluster-admins",
  "id": "cluster-admins"
}

$ ocm --cluster=15uq9gvgj178aumrphhba0nr24j75md0 list users
GROUP               USER
dedicated-admins    jeder@redhat.com
```
a non-CCS cluster without the capability:
```
{
  "kind": "Error",
  "id": "400",
  "href": "/api/clusters_mgmt/v1/errors/400",
  "code": "CLUSTERS-MGMT-400",
  "reason": "'capability.cluster.manage_cluster_admin' is not set for cluster '19dvg7vvmlftufdi57p2474ocio055mv'",
  "operation_id": "1aed7f37-b845-45a0-9544-e2c1f109fbce"
}

$ ocm --cluster=19dvg7vvmlftufdi57p2474ocio055mv list users
GROUP               USER
dedicated-admins    yzhou-nec
dedicated-admins    i-saji
dedicated-admins    takahashi-nes
```
Finally one which has only `dedicated-admins` and even that is empty, in which `list users` didn't even print the header:
```
~ $ ocm get /api/clusters_mgmt/v1/clusters/15uq6splkva07rae2hgih6eph4vs3p8m/groups/clusters-admins
{
  "kind": "Error",
  "id": "404",
  "href": "/api/clusters_mgmt/v1/errors/404",
  "code": "CLUSTERS-MGMT-404",
  "reason": "group with ID 'clusters-admins' not found",
  "operation_id": "0d0989a0-c247-4f47-9c96-d94d21e29770"
}

~ $ ocm get /api/clusters_mgmt/v1/clusters/15uq6splkva07rae2hgih6eph4vs3p8m/groups/dedicated-admins
{
  "kind": "Group",
  "href": "/api/clusters_mgmt/v1/clusters/15uq6splkva07rae2hgih6eph4vs3p8m/groups/dedicated-admins",
  "id": "dedicated-admins"
}

$ ocm list users --cluster=15uq6splkva07rae2hgih6eph4vs3p8m 
```
(not sure if bug or feature, the empty output is caused by `writer.Flush()` being placed inside the users loop so never called.)

## Proposal: Change `list users` so that empty groups are visible.

This will answer both "can I add cluster admins" and "who are they now", and generally will answer which groups I can `ocm add user` to.

One approach is nesting them e.g.:
```
dedicated-admins:
        jeder@redhat.com
cluster-admins:
```
Another is current format but adding a line without user when empty:
```
GROUP               USER
dedicated-admins    jeder@redhat.com
cluster-admins
```

igoihman: This change is in order to maintain backward compatibility and allowing SRE to know if cluster admin is enabled in an easy way. I don't think it's critical having this info in the backend since it's only for visibility and other than this case, there is no valid reason to return this information.

pkg/cluster/describe.go
igoihman: We don't have any validation if cluster admin is enabled in add/delete commands and we are relying on the backend validation. Having said that, we can refactor in a separate change.

PR #223, opened 2021-02-10
Fix empty edit default machine pool

zgalor:
This bug was caused because we use the same func to create and update the nodes builder for a cluster.
It would only update if computes nodes != 0 (which means it was set by the user).
However, the default for edit additional machine pool is -1, because 0 is a valid value.
Since we now use the same `edit machinepool` command for default and additional,
I changed the check to `if computeNodes > 0` which captures both cases.
Related: https://issues.redhat.com/browse/SDA-3709

This also includes a documentation fix for --replicas flag.
Related: https://issues.redhat.com/browse/SDA-3710

### DIFF
--- a/cmd/ocm/edit/machinepool/cmd.go
+++ b/cmd/ocm/edit/machinepool/cmd.go
@@ -61,7 +61,7 @@ func init() {
 		&args.replicas,
 		"replicas",
 		-1,
-		"Restrict application route to direct, private connectivity.",
+		"Count of machines for this machine pool.",
 	)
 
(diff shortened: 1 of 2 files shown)


zgalor:
@cben @vkareh @nimrodshn @igoihman @jhernand PTAL

igoihman (approved):
/lgtm

cben:
LGTM for this situation.  In general, remember that you can use `flags.Changed("flag-name")` to reliably distinguish whether flag was provided or not.

PR #224, opened 2021-02-13
add --dns flag to sshuttle

boranx:
This is to resolve private cluster urls that stopped being resolved after a `dnf upgrade` for fedora33 users.


igoihman (approved):
/lgtm

PR #227, opened 2021-03-02
Add switch for tunnel to use in-continuous IPs

feichashao:
In the current tunnel implementation, tunnel will redirect traffic for the whole subnet of MachineCIDR, ServiceCIDR and PodCIDR. The subnet range is too large, and if the user want to tunnel to multiple clusters, the CIDR has a large chance to conflict.

For SRE, usually we need to tunnel to a target cluster for simple troubleshoot, and having tunnel to API server & console should be enough. Hence, adding a flag --host-only to only tunnel to apiserver & console IPs. 

```
$ ocm tunnel <cluster-name> 
   # /usr/bin/sshuttle --dns --remote sre-user@rh-ssh.xxxx.xxxx.xx.openshiftapps.com 10.aaa.0.0/16 bbb.ccc.0.0/16 ddd.eee.0.0/14

$ ocm tunnel --host-only <cluster-name>
   # /usr/bin/sshuttle --dns --remote sre-user@rh-ssh.xxxx.xxxx.xx.openshiftapps.com 10.aaa.bbb.ccc 10.aaa.ddd.eee 10.aaa.fff.ggg
```
PS. I know `host-only` is a bad name. Let me know if you got a better name.

### DIFF
--- a/cmd/ocm/tunnel/cmd.go
+++ b/cmd/ocm/tunnel/cmd.go
@@ -18,6 +18,8 @@ package tunnel
 
 import (
 	"fmt"
+	"net"
+	"net/url"
 	"os"
 	"os/exec"
 	"regexp"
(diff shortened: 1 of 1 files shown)


mmazur:
This is awesome.

May I suggest making this functionality the default? The only users of this feature are OSD SREs and as far as I can tell the only thing we ever use it for is api and console access. Which means it's a better idea to have the default tunneling option be the non–colliding one.

And the current behavior would then be available under, say, `--subnets` argument, as in `ocm tunnel --subnets CLUSTER`, which sort of reads like a sentence (hey ocm, tunnel subnets for cluster).

feichashao:
@mmazur  Good idea to it default and make the original as --subnets.  --subnets is a good and easy to understand naming.

PR #228, opened 2021-03-05
Set user agent

jhernand:
This patch changes the `ocm` tool so that it explicitly sets the
`User-Agent` header with a value like `OCM-CLI/0.1.47`. This will make
it easier to identify the requests in the logs of the server.


cben:
LGTM

PR #232, opened 2021-04-07
[SDB-1953][SDB-1978] Replace 'allowed' by 'sku_count'

Lir10:


jhernand:
This fails because the `GetTokenExpiry` method was removed from the SDK in a recent change. I will address that in a separate change, then you will be able to rebase this and it should work.

jhernand:
@Lir10 #233 updates the version of the SDK and fixes the problem with the `GetTokenExpiry` function. Once that is merged you should be able to rebase this change.

rawsyntax:
LGTM

PR #233, opened 2021-04-07
Update to SDK 0.1.168

jhernand:
The more relevant change in the new version of the SDK is that the
`GetTokenExpiry` method has been removed. That method should have never
been made public. This patch copies the relevant code to the CLI.


jhernand:
FYI: @Lir10

PR #237, opened 2021-04-21
Add flag to suppress printing of headers

cblecker:
This flag better allows this command to be used in scripting


cblecker:
@igoihman Would you mind reviewing this when you get a moment?

PR #238, opened 2021-04-22
Allow autoscaling non-default machine pool with 0 replicas

zgalor:


zgalor:
@igoihman @nimrodshn @vkareh PTAL

cmd/ocm/edit/machinepool/cmd.go
```diff
@@ -195,10 +196,10 @@ func run(cmd *cobra.Command, argv []string) error {
 	if args.autoscaling.Enabled {
 		asBuilder := cmv1.NewMachinePoolAutoscaling()
 
-		if args.autoscaling.MinReplicas > 0 {
+		if isMinReplicasSet {
```
vkareh, line 199: You should probably check that `isMinReplicaSet && args.autoscaling.MinReplicas >= 0`

zgalor: We don't validate inputs here. We let ocm backend do it and return the error. 
The change was just to allow sending 0 and not treat it as 'not-set'

zgalor:
@vkareh @igoihman @tzvatot this one can already be merged. 

Anyway it depends on the backend to validate input

tzvatot:
LGTM

PR #240, opened 2021-04-27
Adjust column padding for list

cblecker:
Adjust some column padding values to allow the list to display more information with less space padding.


cblecker:
@igoihman Would you mind reviewing this when you get a moment?

PR #241, opened 2021-04-27
remove: revert #244 as we have ocm tunnel -- --dns

boranx:
The flag also can be used as "ocm tunnel ID -- --dns" if needed, so no need to re-define it.
Also, no need to use the cluster-side resolution as the addresses are public.


boranx:
adding: @mmazur

PR #242, opened 2021-04-28
Update to SDK 0.1.168

gshilin-sdb:
The more relevant changes in the new version of the SDK are:
- support of Job Queues
- metrics moved from CS to AMS

### DIFF
--- a/cmd/ocm/cluster/status/status.go
+++ b/cmd/ocm/cluster/status/status.go
@@ -47,27 +47,49 @@ func run(cmd *cobra.Command, argv []string) error {
 	resource := connection.ClustersMgmt().V1().Clusters()
 
 	// Get the resource that manages the cluster that we want to display:
-	clusterResource := resource.Cluster(argv[0])
+	clusterID := argv[0]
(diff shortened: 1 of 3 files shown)


cmd/ocm/cluster/status/status.go
```diff
@@ -58,19 +58,7 @@ func run(cmd *cobra.Command, argv []string) error {
...
-	memUsed := clusterMemory.Used().Value() / 1000000000
-	memTotal := clusterMemory.Total().Value() / 1000000000
-
-	fmt.Printf("State:   %s\n"+
-		"Memory:  %.2f/%.2f used\n"+
-		"CPU:     %.2f/%.2f used\n",
-		cluster.State(),
-		memUsed, memTotal,
-		clusterCPU.Used().Value(), clusterCPU.Total().Value(),
-	)
```
jhernand, line 73: I think it is not OK to just remove these metrics. Can't this be replaced with the metrics obtained from the accounts management service?

gshilin-sdb: Added

PR #248, opened 2021-05-20
Merge search parameter into search query for `list clusters`

jhernand:
Currently when the `--parameter search="..."` flag is used in the `list clusters` command the result is that two `search` query parameters are sent to the server, one containing the value given by the `--parameter` flag and another containing the search query calculated from the rest of the flags. For example, the following command:

```
$ ocm list clusters --managed --parameter search="state is 'ready'"
```

Results in a URL like this:

```
https://api.openshift.com/foo/bar?search=managed+%27t%27&search=state+is+%27ready%27
```

When the server receive that it will just ignore one of the values, so the result will not the expected combination of both search criteria.

To address that this patch changes the `list clusters` command so that it will explicitly combine the search query calculated from other parameters with the values given with the `--parameter` flag. In the above example the result will be a URL like this:

```
https://api.openshift.com/api/clusters_mgmt/v1/clusters?search=%28managed+%3D+%27t%27%29+and+%28state+is+%27ready%27%29
```

Note that this only applies to the `list clusters` command. The `--parameter search="..."` flag will still have the previous meaning in other places.

Related: https://issues.redhat.com/browse/SDA-4201

### DIFF
--- a/cmd/ocm/list/cluster/cmd.go
+++ b/cmd/ocm/list/cluster/cmd.go
@@ -97,14 +97,54 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 	defer connection.Close()
 
-	// Get the client for the resource that manages the collection of clusters:
-	ocmClient := connection.ClustersMgmt().V1().Clusters()
+	// This will contain the terms used to construct the search query:
+	var searchTerms []string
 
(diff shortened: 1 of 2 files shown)


cben:
LGTM!

PR #255, opened 2021-06-09
list versions by channel group

igoihman:


pkg/cluster/versions.go
```diff
@@ -41,14 +42,18 @@ func EnsureOpenshiftVPrefix(v string) string {
...
 // sorted in approximate SemVer order (handling of text parts is somewhat arbitrary).
-func GetEnabledVersions(client *cmv1.Client) (
+func GetEnabledVersions(client *cmv1.Client, channelGroup string) (
 	versions []string, defaultVersion string, err error) {
 	collection := client.Versions()
 	page := 1
 	size := 100
+	filter := "enabled = 'true'"
+	if channelGroup != "" {
+		filter = fmt.Sprintf("%s AND channel_group = '%s'", filter, channelGroup)
```
cben, line 52: With malformed --channel-group his allows injection of arbitrary TSL query, but as this CLI can make arbitrary request anyway and they are validated on the server, I don't see any problem with this.

cben (commented):
LGTM

PR #256, opened 2021-06-09
Add option to encrypt etcd during cluster installation

igoihman:


cben:
LGTM.
Should an inderactive prompt be added for this?

igoihman:
@cben yes, I added the details in Jira

PR #257, opened 2021-06-11
Add ccs_only flag option

pvasant:


pvasant:
@cben @vkareh @tzvatot  PTAL

cmd/ocm/list/region/cmd.go
```diff
@@ -71,18 +108,11 @@ func run(cmd *cobra.Command, argv []string) error {
...
-				region.ID(), true, region.SupportsMultiAZ())
-		}
-	} else {
-		fmt.Fprintf(writer, "ID\t\tON RED HAT INFRA\tON CCS\tSUPPORTS MULTI-AZ\n")
-		for _, region := range regions {
-			fmt.Fprintf(writer, "%s\t\t%v\t%v\t%v\n",
-				region.ID(), region.Enabled(), true, region.SupportsMultiAZ())
-		}
+
+	fmt.Fprintf(writer, "ID\t\tON RED HAT INFRA\tCCS ONLY\tSUPPORTS MULTI-AZ\n")
```
cben, line 112: I'm not sure about displaying the new columns here.
A "CCS ONLY column" is round-about: it's actually about RH infra, and it's negative (CCS ONLY = true means no RH infra, false = can use on RH infra).  And carries an unspoken implication that all regions work on CCS?

My goal in previous design of this table was to answer simple questions of what you can use in which scenarios, without forcing user to understand our exact rules.
So instead of explaining, I had a CCS column filled with "true"; I think giving a independent answer for the 2 scenarios is still a good goal.

Also, having an ON RED HAT INFRA column alongside CCS ONLY sounds confusing to me.  For example if both say "true", isn't that a contradiction?

I'm confused now about what `.enabled` field means (and perhaps I was always confused when I last changed this table?).

- The case when user passes --aws-... flags is more complex.
  We call the same `GetRegions()` function, but the meaning of the result changes, we get a subset relevant to user's AWS account.  I propose that modifying the table headers for that case is still useful.

cmd/ocm/list/region/cmd.go
```diff
@@ -42,25 +43,61 @@ var Cmd = &cobra.Command{
 
 func init() {
 	fs := Cmd.Flags()
-	arguments.AddProviderFlag(fs, &args.provider)
+	fs.StringVar(
+		&args.provider,
+		"provider",
+		"",
+		"Lists the region for the specific cloud provider",
```
cben, line 50: nit: s/region/regions/

cmd/ocm/list/region/cmd.go
```diff
@@ -42,25 +43,61 @@ var Cmd = &cobra.Command{
...
 }
 
 func run(cmd *cobra.Command, argv []string) error {
-	err := arguments.CheckIgnoredCCSFlags(args.ccs)
-	if err != nil {
-		return err
+	ccs := cluster.CCS{}
+	if args.ccs {
+		if args.awsAccessKeyId == "" {
+			return fmt.Errorf("--aws-access-key-id flag is mandatory for --ccs=true")
```
cben, line 81: What if user attempts this with --provider=gcp?
We don't have a similar `available_regions` API for GCP, but that but that asymmetry is not obvious to users.  I propose adding an error for `gcp` and `--ccs` combination, and since you're already tweaking the --help texts to better match this command, change the description of the --ccs flag to explain it's for AWS only.

pvasant:
There was an issue with the rebase and creating another branch was efficient. Closing the PR and creating a new one.

PR #258, opened 2021-06-11
ocm post: pass correct info to ApplyHeaderFlag()

mzazrivec:
The header flags are stored in args.header, not in args.parameter


jhernand (approved):
Looks good, thanks @mzazrivec.

PR #259, opened 2021-06-14
Added ccs_only to cloud regions

pvasant:


pvasant:
@cben  @tzvatot  PTAL 

@cben  sorry i had to close the other PR

cmd/ocm/list/region/cmd.go
```diff
@@ -18,20 +18,21 @@ import (
...
-	ccs      cluster.CCS
+	provider           string
+	ccs                bool
+	awsAccessKeyID     string
+	awsSecretAccessKey string
 }
 
 var Cmd = &cobra.Command{
-	Use:     "regions --provider=CLOUD_PROVIDER [--ccs CCS FLAGS]",
+	Use:     "regions --provider=CLOUD_PROVIDER [--aws --aws-access-key-id --aws-secret-access-key] ",
```
cben, line 35: There is no `--aws` flag, did you mean `--ccs`?

cben:
Sorry for delay, I realized I need to refresh my understanding of what these fields mean in the backend.
Yep, I missed recent https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/3071 (and I was probably confusing parts here with machine types :laughing:) 

Let's see, new logic is:
```go
func (s *Server) validateClusterRegions(
	...
		if !region.Enabled {
			return errors.BadRequest.UserErrorf("Region '%s' is not enabled", *requestedRegion.ID)
		}
		if !s.IsCCS(ccs) && region.CCSOnly {
			return errors.BadRequest.UserErrorf("Region '%s' supports only CCS clusters", *requestedRegion.ID)
		}

		if !region.SupportsMultiAZ && multiAZ {
			return errors.BadRequest.UserErrorf("Region '%s, %s' does not support multiple availability zones",
				region.ID, region.DisplayName)
		}
		return nil
```

cben:
- LGTM to skipping if !Enabled.
- I still think that this:
  ```
  ON RED HAT INFRA  ON CCS 
  true              true          (when ccs_only=false)
  false             true          (when ccs_only=true)
  ```
  might be a clearer display than this:
  ```
  ON RED HAT INFRA    CCS ONLY   
  true                false      
  false               true
  ```
  But that's subjective, if you prefer the latter, also OK.

cben:
@igoihman LGTM, please review/merge.
We need to get a new version out as API changed and previous versions will misleadingly show all clusters as working on both infras.

cben:
(I also tested new code against new data coming from stage and it works right.)

Issue #226, opened 2021-02-23
Flag --columns api.url returns incorrect API url


apodhrad:
The following command works correctly
```
ocm list cluster --parameter search="name='osd4-153'"
```
and returns correct ID, NAME, API URL and other columns.
Where API URL is `https://api.osd4-153.fneg.s1.devshift.org:6443`

But when I request a specific column by
```
ocm list cluster --parameter search="name='osd4-153'" --columns api.url
```
then I receive only shorter API URL, namely `https://api.osd4-153.fneg.s1.dev`

OCM Version: 0.1.47
OS: Fedora 33 x86_64

cben:
Yes, columns are currently truncated.
You can control width e.g. `--columns api.url --padding=100`.
But I agree, silently truncating is a bug / mis-feature.  Better options when a field is wider than allocated column width:

1. at least could indicate it's truncated, e.g. with ellipsis char?
2. break alignment but include full value!
  - con: output is no longer machine-parsable.  Meh.  Knowing expected positions e.g. `cut --characters=40-59` is fragile anyway. And who needs machine-parsable format when the data is truncated?
  - con: truncating _is_ useful to fit an "overview" of data into terminal width.  However, piping to `less -S` and scrolling horizontally is more flexible (you need this anyway for default columns).
3. use Tab characters between columns; pad with spaces (before the tab) to align to given width, but allow wider texts.
4. increase column width to fit!  However, this requires buffering the whole output; `ocm list cluster` is getting a "page" at a time from the API and currently showing them as they come, buffering it all can take a _long_ time so is not good default.

IMHO (3) is best option.  
+ Pro: More parsable than current output.  
+ Pro: Users that want (4) and willing to wait for full output can pipe to filters like `| column --table --separator=$'\t'` or https://github.com/BurntSushi/tabwriter.

cben:
- I see only `ocm list` commands do pagination but have no json output.  We should add `--jsonlines` or something like that, independently of what we decide in this issue.

cben:
Responding to a ideas by @T0MASD from related issue on [internal tracker](https://issues.redhat.com/browse/SDA-4246):

> would it make sense to never pad the data when --columns flag is used? 

You mean always have fixed 1 or 2 spaces between columns, and not even attempt alignment?  It's doable, but IMHO would be a worse default.
- Pro: doesn't pretend columns are reliable when they really are not.
[...]

PR #262, opened 2021-06-21
Add table printer

jhernand:
Add table printer

This patch adds a new `output.Table` type that simplifies writing
tabular ouptut. It is created from the `output.Printer` type introduced
in a previous patch:

```go
// Create the output table:
table, err := printer.NewTable().
        Name("clusters").
        Columns(args.columns).
        Build(ctx)
if err != nil {
        return err
}
```

When a table like this is created it reads the descriptions of the
columns from the corresponding file in the `pkg/output/tables`
directory. In this case it will read the `clusters.yaml` file. That file
should contain the descriptions of the columns of that table, in
particular it should contain the preferred width of the columns. For
example:

```yaml
columns:
- name: id
  header: ID
  width: 32
- name: name
  header: NAME
  width: 28
  ...
- name: external_id
  header: EXTERNAL ID
  width: 36
```

In this example the `external_id` column is defined with a preferred
width of 36 charaters. The table object will use that information to
format the output accordingly. For example, a command like this:

```
$ ocm list clusters --columns id,external_id,name --managed
```

Will result in the following output:

```
ID                                EXTERNAL_ID                   NAME
1i6anlpc6278j4ok5nbp8ik69dl6h53j  9f2c1195-1f1b-4780-ae09-00f6  osde2e-hwlrn
1ku25di05s69u46bbnc0abihl9ig2e9a  155beb89-feec-4246-aab9-1592  osde2e-hxhfh
1lfklioadcqr5oeitfokqkavq4gir25l  e2eb5d92-f0d7-4077-b7b3-a2d8  osde2e-kp4sk
1lfklji54fhu3b64us3a62d9t9bju2r8  NONE                          osde2e-wfw97
1lfkljjopet5vkpnks33j65kb7pilao3  3e567572-f4fb-49cf-b25e-3196  osde2e-o9z8x
...
```

Note that this patch only modifies the `list clusters` command. Other
commands that write tables are still using the previous infrastructure.
That will be addressed in separate patches.

Related: https://issues.redhat.com/browse/SDA-4246

### DIFF
--- a/pkg/output/table.go
+++ b/pkg/output/table.go
@@ -1,0 +1,327 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// This file contains the code that writes generates tabular output.
+
+//go:generate go-bindata -pkg output tables
+
+package output
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/openshift-online/ocm-cli/pkg/data"
+	"gopkg.in/yaml.v3"
+)
+
+// TableBuilder contains the data and logic needed to create a new output table.
+type TableBuilder struct {
+	printer *Printer
+	name    string
+	specs   []string
+	digger  *data.Digger
+}
+
+// Table contains the data and logic needed to write tabular output.
+type Table struct {
+	printer *Printer
+	name    string
+	columns []*Column
+	digger  *data.Digger
+}
+
+// tableYAML is used to load a table description from a YAML document.
+type tableYAML struct {
+	Columns []*columnYAML `yaml:"columns"`
+}
+
+// Column contains the data and logic needed to write columns.
+type Column struct {
+	name   string
+	header string
+	width  int
+}
+
+// columnYAML is used to load a column description from a YAML document.
+type columnYAML struct {
+	Name   *string `yaml:"name"`
+	Header *string `yaml:"header"`
+	Width  *int    `yaml:"width"`
+}
+
+// NewTable creates a new builder that can then be used to configure and create a table.
+func (p *Printer) NewTable() *TableBuilder {
+	return &TableBuilder{
+		printer: p,
+	}
+}
+
+// Name sets the name of the table. This is mandatory.
+func (b *TableBuilder) Name(value string) *TableBuilder {
+	b.name = value
+	return b
+}
+
+// Column adds a column to the table. The spec can be a single column identifier or a set of comma
+// separated column identifiers.
+func (b *TableBuilder) Column(spec string) *TableBuilder {
+	b.specs = append(b.specs, spec)
+	return b
+}
+
+// Columns adds a collection of columns to the table. Each spec can be a single column identifier or
+// a set of comman separated column identifiers.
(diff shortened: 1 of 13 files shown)


pkg/output/tables/clusters.yaml
```diff
@@ -0,0 +1,44 @@
...
+    width: 14
+  region.id:
+    header: REGION
+    width: 14
+  state:
+    header: STATE
+    width: 13
+  external_id:
+    header: EXTERNAL_ID
+    width: 36
```
T0MASD, line 44:
```suggestion
    width: 36

```

jhernand: Thanks @T0MASD, I will add this in the next version of the patch.

cben:
Not a full review yet, some high-level thoughts from skimming:

- The .yaml files get bundled into the binary, right?
  Perhaps it would be simpler to describe the data in Go maps & structs, rather than parse it on the fly.  Fine either way.

- I'm tempted in the future to process dynamic-columns requests by deserializing JSON into untyped data (`Items []map[string]interface{}`) rather than reflection on SDK methods.
  This is definitely out of scope now; but to that end, I like the separation of concerns of `Digger` :+1:.

  - Conversely, if you feel that using SDK types + reflection is a better way, consider in future incorporating something like Digger into the SDK itself.  [out of scope, fine to prototype it here for now.]

- Are `methodCache`, `fieldCache` premature optimization?
  - Since we know the set of desired columns before we start processing, you could pre-populate these on construction rather than dynamic thread-safe add-if-not-there.  Would that simplify or complicate things?

pkg/output/table.go
```diff
@@ -0,0 +1,322 @@
...
+		if columnValue != nil {
+			columnText = fmt.Sprintf("%v", columnValue)
+		} else {
+			columnText = "NONE"
+		}
+		actualWidth := len(columnText)
+		desiredWidth := t.columns[i].width
+		switch {
+		case actualWidth > desiredWidth:
+			rowBuffer.WriteString(columnText[0:desiredWidth])
```
cben, line 289: This still silently truncates texts that don't fit.  
IMHO that does not solve #226 / https://issues.redhat.com/browse/SDA-4246, the fundanmental question there is not "good defaults" but "what to do when it didn't fit".

However, it's an improvement on current one-width-fits-all behavior, and improving the truncation behavior is an orthogonal change that will be easier once all commands use `printer`, so :+1:

jhernand:
The main reason for using the YAML files instead of Go types is that it isn't then necessary to have a place where all those definitions are explicitly listed, something like this:

```go
// In table_index.go maybe?
var tableIndex map[string]*TableDescription = map[string]*TableDescription{
        "clusters": clustersTable,
        "versions": versionsTable,
        ...
}
```

It isn't a strong reason probably.

Using JSON as an intermediate format to extract table data seems reasonable. However it would not work for objects that can't be serialized to JSON. I don't see a need to restrict ourselves only to objects that can be serialized.

I'd say that not using the caches for methods and fields would be a premature performance decrease. But truth is that I didn't check what is the difference in performance.

Having to explicitly specify the methods used to extract each column in advance would be really inconvenient, in my opinion.

pkg/output/table.go
jhernand: One of the things that we can do once we have metadata for tables/columns is introduce rules to decide what to do when they don't fit. For example, we could have something like this:

```yaml
columns:
- name: api.url
  trim: false
```

That would tell the CLI that the "api.url" column should never be trimmed. It could then use this information to move the column to the end of the table, and never trim it. Obviously this can't be done if there are multiple columns with `trim: false`, but the point is that it makes it possible to make the display logic much more intelligent.

Another possibility would be to at least mark those columns to explicitly indicate that they have been trimmed:

```yaml
columns:
- name: api.url
  ellipsis: true
```

That hypothetical configuration would mean that the value should be terminaled with `…` to visually indicate that it has been trimmed:

```
$ ocm list clusters --columns id,api.url,name
ID                                API URL                       NAME
1i6anlpc6278j4oa5nbp8ik69dl6h53j  https://api.osde2e-hcltn.3a…  osde2e-hcltn
```

So, I agree that this doesn't solve all the problems, it is just an starting point.

PR #263, opened 2021-06-22
Revert "Use `less` to page cluster list"

jhernand:
This reverts commit 794f94891b9c6253d7a96f3145232a2e6beaa097 because the
new default behaviour is breaking the workflow of some users. In
particular uses have commented that the fact that _less_ clears the
screen after finishing isn't convenient.

Related: https://issues.redhat.com/browse/SDA-4280


jhernand:
FYI: @cblecker @igoihman

PR #267, opened 2021-06-23
[WIP] Use pager to list clusters

jhernand:
Use pager command for `list clusters`
    
This patch changes the `list clusters` command so that it uses the
a pager command (if available) to display the results page by page.

This will not be enabled by default. To enable it set the `pager` config
variable to the command that you wish to use. For example, to use the
`less` command:
    
```
$ ocm config set pager less
```


jhernand:
Replaced by #274.

PR #272, opened 2021-06-29
Don't remove config file on logout

jhernand:
Currently the `ocm logout` command removes the `$HOME/.ocm.json`
configuration file.  his makes it difficult to have settings that
aren't related to authentication. For example, we want to introduce a
setting to enable/disable paging, and we would like to have that
persisted in the configuration file so that the user doesn't need to
change it after each logout. To allow that this patch changes the
`logout` command so that instead of removing the configuration file it
removes all the login related settings.


cben:
LGTM, very nice :clap:

PR #286, opened 2021-07-09
Use subscription API to query cluster id

wanghaoran1988:
Use subscription API to query cluster id when cluster name or external ID provided.
More context: https://issues.redhat.com/browse/SDB-2155, https://issues.redhat.com/browse/SDB-1706

### DIFF
--- a/cmd/ocm/describe/cluster/cmd.go
+++ b/cmd/ocm/describe/cluster/cmd.go
@@ -25,6 +25,7 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
+	c "github.com/openshift-online/ocm-cli/pkg/cluster"
 	clusterpkg "github.com/openshift-online/ocm-cli/pkg/cluster"
 	"github.com/openshift-online/ocm-cli/pkg/dump"
 	"github.com/openshift-online/ocm-cli/pkg/ocm"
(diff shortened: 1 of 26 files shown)


cmd/ocm/describe/cluster/cmd.go
```diff
@@ -93,25 +93,30 @@ func run(cmd *cobra.Command, argv []string) error {
...
 		os.Exit(1)
 	}
-	cluster := clusters[0]
-
+	clusterID := subs[0].ClusterID()
+	clusterResponse, err := connection.ClustersMgmt().V1().Clusters().Cluster(clusterID).Get().Send()
+	if err != nil {
+		return fmt.Errorf("Can't retrieve cluster for key '%s': %v", key, err)
+	}
+	cluster := clusterResponse.Body()
```
jhernand, line 119: Can/should this ^ code be replaced by `c.GetCluster(connection, key)`?

jhernand (commented):
Looks good to me. Just a minor comment about re-using `c.GetCluster` in case you want to do that before merging.

jhernand:
By the way, there is a small test framework in place now. Would be nice if you can add a test to verify that this is working correctly. The tests are here:

https://github.com/openshift-online/ocm-cli/blob/master/tests

For example:

https://github.com/openshift-online/ocm-cli/blob/master/tests/list_clusters_test.go

jhernand (approved):
Excellent, thanks @wanghaoran1988!

PR #289, opened 2021-07-13
feat(login): allow for empty OCM_CONFIG

georgettica:
- needed if the file is overridden with "export OCM_CONFIG=$(mktemp)"
- go.mod and go.sum updated via the `make` target


georgettica:
the error I get without the fix is:
```
ocm login --token $TOKEN
Error: Cant load config file: cant parse config file $FILE: unexpected end of JSON input
```

jhernand (approved):
Looks good, thanks @georgettica.

georgettica:
could I get pinged when the next release is being cut?
I have code that depends on this change

jhernand:
You can use the GitHub "watch" button to configure the notifications that you want to receive. In particular you can configure it to be notified when a new release is made.

PR #291, opened 2021-07-14
Fix cluster lookup

jhernand:
This patch changes the cluster lookup logic so that that it will search
for subscriptions in `Reserved` and `Active` state, and so that it will
look in the clusters collection when there are no matching results in
the subscriptions collection.

Related: https://issues.redhat.com/browse/SDA-4458

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -110,31 +110,95 @@ func IsValidClusterKey(clusterKey string) bool {
 	return clusterKeyRE.MatchString(clusterKey)
 }
 
-func GetCluster(connection *sdk.Connection, clusterKey string) (*cmv1.Cluster, error) {
-	search := fmt.Sprintf("(display_name = '%s' "+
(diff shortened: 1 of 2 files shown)


pkg/cluster/cluster.go
```diff
@@ -110,31 +110,95 @@ func IsValidClusterKey(clusterKey string) bool {
...
+
+	// Try to find a matching subscription:
+	subsSearch := fmt.Sprintf(
+		"(display_name = '%s' or cluster_id = '%s' or external_cluster_id = '%s') and "+
+			"status in ('Reserved', 'Active')",
+		key, key, key,
+	)
+	subsListResponse, err := subsResource.List().
+		Search(subsSearch).
+		Size(1).
```
tiwillia, line 126: If the key provided is the display name and multiple subscriptions match that display name, this will just return whichever is the first one provided in the API. That is probably fine behavior, but just something we should take note of.

PR #293, opened 2021-07-19
Add htpasswd IDP

NautiluX:
Allow creation of an HTPasswd IDP


zgalor:
@pvasant added that feature in CS, I want her to review it as well.

cmd/ocm/create/idp/htpasswd.go
```diff
@@ -0,0 +1,63 @@
...
+		prompt := &survey.Password{
+			Message: "Enter password:",
+		}
+		err = survey.AskOne(prompt, &password)
+		if err != nil {
+			return idpBuilder, errors.New("Expected a password")
+		}
+	}
+
+	// Create Google IDP
```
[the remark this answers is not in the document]
NautiluX: thanks, fixed

NautiluX:
@georgettica the help of the command should show the options, where else do you think I should add a description?

georgettica:
here:
https://github.com/openshift-online/ocm-cli/pull/293/files#diff-75ceaa3482cb1ab5de68a6b4b0a584863e4b9e31ac97b46adb8b1473aec2754eL69-L72

NautiluX:
> here:
> https://github.com/openshift-online/ocm-cli/pull/293/files#diff-75ceaa3482cb1ab5de68a6b4b0a584863e4b9e31ac97b46adb8b1473aec2754eL69-L72

That help contains only an example for a GitHub IDP. To me the purpose seems to be to show an example of one of the most often used IDP types.
The supported IDP types are listed at the [-type flag](https://github.com/openshift-online/ocm-cli/blob/90e6dd69c10143c0ac48b0ebad3e498a5a49951d/cmd/ocm/create/idp/cmd.go#L96).

georgettica:
okay :)
at least the OSD docs team need to know of this?

georgettica:
/lgtm

after discussion

cmd/ocm/create/idp/htpasswd.go
```diff
@@ -0,0 +1,64 @@
...
+		Password(password)
+
+	// Create new IDP with HTPasswd provider
+	idpBuilder.
+		Type("HTPasswdIdentityProvider"). // FIXME: ocm-api-model has the wrong enum values
+		Name(idpName).
+		MappingMethod(cmv1.IdentityProviderMappingMethod(args.mappingMethod)).
+		Htpasswd(htpasswdIDP)
+
+	fmt.Println("Securely store your username and password.")
```
pvasant, line 61: @NautiluX  Thanks. Overall LGTM.  I would add these after the IDP is actually created successfully so it doesn't get lost in the trace 

https://github.com/openshift-online/ocm-cli/blob/85ebe2ebca0465b978e98eb36a2ae3f917e6b78c/cmd/ocm/create/idp/cmd.go#L347

pvasant, line 61: Thanks, @NautiluX . Overall LGTM. I would add this after IDP is successfully created so it doesn't get lost in the trace 

https://github.com/openshift-online/ocm-cli/blob/85ebe2ebca0465b978e98eb36a2ae3f917e6b78c/cmd/ocm/create/idp/cmd.go#L347

Can we expand IDP to either identity provider or this is what we have currently in ROSA 
`
Please securely store this generated password. If you lose this password you can delete and recreate the cluster admin user.`

NautiluX: Do you suggest to only allow generated passwords rather than getting one as input? The UI allows to specify a password, so I thought it would be okay to allow it as input. We could use a generated one if none is entered?
For the output, the current implementation doesn't allow IDP-specific output after the creation, so I added it before, I think doing that without a `if htpasswd` would involve quite some refactoring. Do you think the order of log lines is worth the refactoring?

pvasant: @NautiluX  no :)  I just wanted to move the log lines

fmt.Println("Securely store your username and password.")
fmt.Println("If you lose these credentials, you will have to delete and recreate the IDP.")

 after we successfully created the idp in  cmd/ocm/create/idp/cmd.go#L353

NautiluX: uhm ok, now i added the generation option... You like it? :D
I'm not sure I like how htpasswd is handled differently now regarding the additional log lines.

PR #294, opened 2021-07-19
Update login URL for upcoming move to console.redhat.com

cben:
The new URL already works so it's safe to merge & release this, and since users take time to upgrade the CLI, better sooner IMHO.

related to https://issues.redhat.com/browse/OSDOCS-2290


igoihman (approved):
/lgtm

PR #296, opened 2021-07-20
Clarify `ocm completion` help

cben:
It's regrettable for new users that running `ocm completion` with no args dumps bash script to stdout — ideally it'd show the help and require an explicit shell type e.g. `ocm completion bash`.
I kept the shell type optional to not break "existing users", but does that matter to anyone?


igoihman (approved):
/lgtm

PR #302, opened 2021-08-05
config: Use standard config path for ocm.json

vkareh:
If the legacy ~/.ocm.json file already exists, it continues using that.
Otherwise it prefers the standard XDG config directory.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -96,6 +96,11 @@ func Save(cfg *Config) error {
 	if err != nil {
 		return err
 	}
+	dir := filepath.Dir(file)
+	err = os.MkdirAll(dir, os.FileMode(0755))
+	if err != nil {
+		return fmt.Errorf("can't create directory %s: %v", dir, err)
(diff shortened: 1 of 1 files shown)


vkareh:
relevant: https://github.com/openshift/rosa/pull/417

pkg/config/config.go
```diff
@@ -107,17 +112,34 @@ func Save(cfg *Config) error {
...
+	// Determine home directory to use for the legacy file path
+	home, err := homedir.Dir()
+	if err != nil {
+		return "", err
+	}
+
+	path = filepath.Join(home, ".ocm.json")
+
+	_, err = os.Stat(path)
+	if os.IsNotExist(err) {
```
igoihman, line 132: nit: 
```if _, err := os.Stat("/path/to/whatever"); os.IsNotExist(err) {
// path/to/whatever does not exist
}```

igoihman (approved):
/lgtm

Issue #295, opened 2021-07-19
Using of JQ leads to the loss on int64 numbers


petli-openshift:
I am working to add a new API that one of the field is int64. I noticed there's loss using `ocm` v.s. using `curl`,
```
echo '{"type":"sca", "arch":"x86_64"}' | ocm post http://localhost:8500/api/accounts_mgmt/v1/certificates
------ response ------
"id": 2377845297775806000,

curl http://localhost:8500/api/accounts_mgmt/v1/certificates -d '{"type":"sca", "arch":"x86_64"}'
------ response ------
"id":2377845297775805901,
```

After looking into the code, I think the cause is integration of jq (as a terminal highlighter?): https://github.com/openshift-online/ocm-cli/blob/90e6dd69c10143c0ac48b0ebad3e498a5a49951d/pkg/dump/dump.go#L31-L32

The loss with large numbers is a known issue to jq. Here's one of the most recent discussion: https://github.com/stedolan/jq/issues/1959
```
$ echo 4418489049307232906 | jq
4418489049307233000
```
This is really an edge case. I don't expect removing the jq integration because of this. However, will you consider adding an option for turning it off? so for rare cases like this, if a person does need the raw output, he can specify a flag to skip jq.

cben:
Ooh, great find ❤️ ☹️ 

For an immediate workaround, pipe output to `| cat` this will supress colorization by `jq`.

We can look at other ways to colorize in-process.  But the benefit will go away if user pipes to jq, which realistically is a common wya to use ocm.
Javascript also has trouble with full int64 range.
Maybe we can re-think the API...

cben:
Can you make that id a string?  That will help various clients...

IEEE doubles retain integer precision up to 2**53.  Is 2**64 really the range you need, or might they grow even bigger?
(BTW, is it opaque to the user - as most our fields named `id`?)

petli-openshift:
unfortunately, it's not us that create these. We merely pass the data through from Red Hat Subscription Management to the client. This will unlikely cause any issue since the API will be mostly used by a specific client app, which does not use ocm-cli.
Just wanted to give a heads up here. There's really no need change anything because of it. It could be considered with other improvements related to the jq usage in ocm.

jhernand:
This issue is addressed in #315.

georgettica:
Added a note if a jq alternative in #315 , which is https://github.com/itchyny/gojq

I hope the extra info helps

jhernand:
[...]

PR #315, opened 2021-08-30
Don't use `jq` for color output

jhernand:
This patch changes the tool so that it doesn't rely on the `jq` tool in
order to produce color output.

The main reason to do this is to prevent incorrect handling of long
integers.

Related: https://github.com/openshift-online/ocm-cli/issues/295

### DIFF
--- a/pkg/dump/dump.go
+++ b/pkg/dump/dump.go
@@ -18,18 +18,17 @@ limitations under the License.
 package dump
 
 import (
-	"bytes"
 	"encoding/json"
 	"io"
-	"os"
-	"os/exec"
 
+	"github.com/nwidger/jsoncolor"
+	"github.com/openshift-online/ocm-cli/pkg/output"
 	"gitlab.com/c0b/go-ordered-json"
 )
 
 // Pretty dumps the given data to the given stream so that it looks pretty. If the data is a valid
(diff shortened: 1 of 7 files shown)


pkg/dump/dump.go
```diff
@@ -57,46 +68,41 @@ func Simple(stream io.Writer, body []byte) error {
...
+func dumpBytes(stream io.Writer, data []byte) error {
+	_, err := stream.Write(data)
+	if err != nil {
+		return err
+	}
+	_, err = stream.Write([]byte("\n"))
+	return err
+}
+
+func isTerminal(writer io.Writer) bool {
```
cben, line 101: We already use `github.com/mattn/go-isatty` in another place, let's use one mevhanism for both?

cben (approved):
LGTM 🌈

pkg/dump/dump.go
jhernand: That makes sense. I will do it.

jhernand:
Changed to use the same TTY check library in all places.

georgettica:
I just saw this change, and I wanted to note a tool I found recently that might be useful here or in general

https://github.com/itchyny/gojq

Which also supports working as a library and works 100% in go

jhernand:
Very interesting, thanks @georgettica, I will take that library into account in the future.

PR #316, opened 2021-09-08
describe: Show provisioning error code and message

vkareh:


gregsheremeta:
logic looks right to me

who can merge? :)

PR #323, opened 2021-09-27
Add ServiceLog count to describe cluster

cblecker:
This upgrades a couple dependencies, and then adds the field to count the SLs for the cluster (total, and total in last week).


gregsheremeta:
is there a corresponding Jira card? Just curious what the use case is, especially for the in-last-week summary part

cblecker:
@gregsheremeta There isn't a JIRA card. Use case is SREs understanding if there are active service logs that have been sent.. a hint to go review the SL

igoihman:
closing this one, as this is not info we'd like to expose to customers

igoihman:
closing this one, as this is not info we'd like to expose to customers

PR #332, opened 2021-11-23
Update README.adoc

adlerOri:
- Fix typo in **Multiple Concurrent Logins** section.
- Fix query parameter of ``ocm get /api/clusters_mgmt/v1/cloud_providers/aws/regions``.
  At the moment, the command returns error 400 - *The name is not a valid field name*.


adlerOri:
@zgalor PTAL

PR #333, opened 2021-11-29
Fix bug - return exit code 1 when user creating is failed

adlerOri:
Change the **cmd.go** file of the ``create user`` command to return exit
code 1 when user creating is failed with an existing username.
After the changes will be merged:
- When creating a user is failed the exit code will be 1.
- When creating multiple users, if one failed the exit status code
will be 1.

Solve Bug: SDA-4804

### DIFF
--- a/cmd/ocm/create/user/cmd.go
+++ b/cmd/ocm/create/user/cmd.go
@@ -107,6 +107,8 @@ func run(cmd *cobra.Command, argv []string) error {
 		return fmt.Errorf("Group '%s' in cluster '%s' doesn't exist", args.group, clusterKey)
 	}
 
+	failedToAddUser := false
+	var usersFailed []string
 	for _, username := range strings.Split(users, ",") {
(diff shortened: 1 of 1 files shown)


adlerOri:
@zgalor @tzvatot PTAL

cmd/ocm/create/user/cmd.go
```diff
@@ -121,10 +123,15 @@ func run(cmd *cobra.Command, argv []string) error {
 			Send()
 		if err != nil {
 			fmt.Printf("Failed to add '%s' user '%s' to cluster '%s': %v\n", args.group, username, clusterKey, err)
+			failedToAddUser, addUserErr = true, err
```
zgalor, line 126: If there will be multiple failures the last one will override the ones before.
Try to return a list of the users that failed:

```
return fmt.Errorf("Failed to create the following users in group '%s': user1, user4", args.group"
```

zgalor:
LGTM

PR #341, opened 2021-12-08
Add STS to describe cluster output

fahlmant:
Shows STS true or false based on the existence of STS information for the cluster.

### DIFF
--- a/pkg/cluster/describe.go
+++ b/pkg/cluster/describe.go
@@ -126,8 +126,12 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster) 
 	}
 
 	privateLinkEnabled := false
+	stsEnabled := false
 	if cluster.CloudProvider().ID() == ProviderAWS && cluster.AWS() != nil {
 		privateLinkEnabled = cluster.AWS().PrivateLink()
+		if cluster.AWS().STS().RoleARN() != "" {
+			stsEnabled = true
+		}
 	}
 
 	// Print short cluster description:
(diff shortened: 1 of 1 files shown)


pkg/cluster/describe.go
```diff
@@ -130,6 +130,11 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
 		privateLinkEnabled = cluster.AWS().PrivateLink()
 	}
 
+	stsEnabled := false
+	if cluster.AWS().STS() != nil {
```
pvasant, line 134: Do we need nil validation for 

 if cluster.AWS()!=nil  &&  cluster.AWS().STS() != nil

fahlmant: Good point, let me put it in the same block as the privateLink check above

vkareh: No, this should be fine, as the types are initialized. However, this also means that checking for `STS != nil` is not sufficient, you'll need something like:
```go
if cluster.AWS().STS().RoleARN() != "" {
  stsEnabled = true
}
```

pkg/cluster/describe.go
```diff
@@ -126,8 +126,12 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
 	}
 
 	privateLinkEnabled := false
+	stsEnabled := false
 	if cluster.CloudProvider().ID() == ProviderAWS && cluster.AWS() != nil {
 		privateLinkEnabled = cluster.AWS().PrivateLink()
+		if cluster.AWS().STS() != nil {
```
vkareh, line 132: Use `cluster.AWS().STS().RoleARN() != ""`

fahlmant: Makes sense, will do

PR #344, opened 2022-01-04
It should successful when create cluster with subnets via ocm-cli

zvikorn:

https://issues.redhat.com/browse/SDA-5260

bug: OCM-CLI sent same availability zone twice instead of one


jhernand (approved):
Looks good to me.

PR #349, opened 2022-01-12
OCM-CLI: Bad Request when creating a cluster without subnet IDs

zvikorn:

https://issues.redhat.com/browse/SDA-5308


vkareh:
Thanks!

/lgtm

PR #350, opened 2022-01-17
OCM-CLI: Support cluster-wide proxy during cluster creation and update

zvikorn:

https://issues.redhat.com/browse/SDA-5004
https://issues.redhat.com/browse/SDA-5204

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -31,6 +31,7 @@ import (
 	c "github.com/openshift-online/ocm-cli/pkg/cluster"
 	"github.com/openshift-online/ocm-cli/pkg/ocm"
 	"github.com/openshift-online/ocm-cli/pkg/provider"
+	"github.com/openshift-online/ocm-cli/pkg/utils"
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift/rosa/pkg/interactive"
@@ -56,6 +57,8 @@ var args struct {
 	private               bool
 	multiAZ               bool
 	ccs                   c.CCS
+	existingVPC           c.ExistingVPC
+	clusterWideProxy      c.ClusterWideProxy
 	gcpServiceAccountFile arguments.FilePath
 	etcdEncryption        bool
 
@@ -63,7 +66,6 @@ var args struct {
 	computeMachineType string
 	computeNodes       int
 	autoscaling        c.Autoscaling
-	byovpc             c.BYOVPC
 
 	// Networking options
 	hostPrefix  int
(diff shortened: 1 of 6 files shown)


cmd/ocm/create/cluster/cmd.go
```diff
@@ -56,14 +58,15 @@ var args struct {
 	private               bool
 	multiAZ               bool
 	ccs                   c.CCS
+	byovpc                c.BYOVPC
```
igoihman, line 61: please avoid using `byo-vpc` terminology and use existing vpc instead

zvikorn: done

cmd/ocm/create/cluster/cmd.go
```diff
@@ -511,8 +525,19 @@ func run(cmd *cobra.Command, argv []string) error {
...
+}
+
+func wasClusterWideProxyReceived() bool {
+	return (args.clusterWideProxy.HTTPProxy != nil && *args.clusterWideProxy.HTTPProxy != "") ||
+		(args.clusterWideProxy.HTTPSProxy != nil && *args.clusterWideProxy.HTTPSProxy != "") ||
+		(args.clusterWideProxy.AdditionalTrustBundle != nil && *args.clusterWideProxy.AdditionalTrustBundle != "")
+}
+
+func isProxyEmpty() bool {
+	return *args.clusterWideProxy.HTTPProxy == "" && *args.clusterWideProxy.HTTPSProxy == "" &&
```
ciaranRoche, line 539: Should there be a nil check on these before we dereference them?

zvikorn: on the other hand, I think this should be left as is. As I said, there is no way they are nil. If they are, something is wrong.

pkg/arguments/arguments.go
```diff
@@ -157,14 +157,14 @@ func CheckIgnoredCCSFlags(ccs cluster.CCS) error {
 	return nil
 }
 
-func AddBYOVPCFlags(fs *pflag.FlagSet, value *cluster.BYOVPC) {
+func AddExistingVPCFlags(fs *pflag.FlagSet, value *cluster.ExistingVPC) {
 	fs.BoolVar(
 		&value.Enabled,
-		"byo-vpc",
+		"use-existing-vpc",
```
pvasant, line 163: We use "Install into an existing VPC" in ROSA if we want to keep it consistent across all the cli we manage

cmd/ocm/create/cluster/cmd.go
```diff
@@ -533,19 +558,115 @@ func promptName(argv []string) error {
 	return fmt.Errorf("A cluster name must be specified")
 }
 
-func promptBYOVPC(fs *pflag.FlagSet, connection *sdk.Connection, cmd *cobra.Command) error {
+func shouldPromptForClusterWideProxy(proxy c.ClusterWideProxy, ccs c.CCS, interactive bool) bool {
```
pvasant, line 561: we can just change it to  promptForClusterWideProxy

cmd/ocm/create/cluster/cmd.go
```diff
@@ -511,8 +525,19 @@ func run(cmd *cobra.Command, argv []string) error {
 	return nil
 }
 
-func shouldPromptForBYOVPC(areSubnetsProvided bool) bool {
-	return !args.byovpc.Enabled && args.ccs.Enabled && !areSubnetsProvided && args.interactive
+func shouldPromptForExistingVPC(areSubnetsProvided bool) bool {
+	return !args.existingVPC.Enabled && args.ccs.Enabled && !areSubnetsProvided && args.interactive
+}
+
+func wasClusterWideProxyReceived() bool {
```
pvasant, line 532: we can just change it to  clusterWideProxyReceived

zvikorn: I rather leave it as is, since the function returns boolean. The function answers the question - "was cluster-wide proxy received"

cmd/ocm/create/cluster/cmd.go
zvikorn: I rather leave it as is, since the function returns boolean. The function answers the question - "should we prompt the user with cluster-wide proxy option"

jhernand:
Looks good to me.

PR #357, opened 2022-01-28
Adds more URLs to expander

markturansky:
* Adds "quota_cost" URL.
* Changes ID from simple append to string format
* adds explicit IDs to all tests


markturansky:
@rawsyntax  @jhernand PTAL

rawsyntax:
LGTM

PR #358, opened 2022-01-29
The attribute name in error message should be same with the correct proxy attribute in body

zvikorn:

https://issues.redhat.com/browse/SDA-5394


oriAdler:
LGTM
Should we also add the word 'additional' to the error message on line 53?

zvikorn:
@oriAdler good catch.Thanks!

PR #360, opened 2022-01-29
There should be error message when create a ccs cluster with no exiting vpc but set proxy

zvikorn:

https://issues.redhat.com/browse/SDA-5393


oriAdler:
LGTM

PR #372, opened 2022-02-22
Fix HTML escaping

tzvatot:
@oisakov @jhernand 

Related: [SDA-5048](https://issues.redhat.com/browse/SDA-5048)

### DIFF
--- a/pkg/dump/dump.go
+++ b/pkg/dump/dump.go
@@ -46,6 +46,7 @@ func Pretty(stream io.Writer, body []byte) error {
 
 func dumpColor(stream io.Writer, data interface{}) error {
 	encoder := jsoncolor.NewEncoder(stream)
+	encoder.SetEscapeHTML(false)
 	encoder.SetIndent("", "  ")
 	return encoder.Encode(data)
 }


jhernand (commented):
I am afraid this is not going to work completely. There are other places in this file where we create encoders, those also need to be changed. In addition we use an ordered map to make sure that the order of the fields in the output isn't different to the order of the fields received from the server. That type implements the `MarshalJSON` method, and internally it uses the `encoding/json` package in a way that makes it impossible to use the `encoder.SetEscapeHTML` method. So if we really need to address this completely we will need to stop using that ordered map (and accept that the output may be in a different order than the data received from the server) or else look for an alternative solution.

tzvatot:
> I am afraid this is not going to work completely. There are other places in this file where we create encoders, those also need to be changed. In addition we use an ordered map to make sure that the order of the fields in the output isn't different to the order of the fields received from the server. That type implements the `MarshalJSON` method, and internally it uses the `encoding/json` package in a way that makes it impossible to use the `encoder.SetEscapeHTML` method. So if we really need to address this completely we will need to stop using that ordered map (and accept that the output may be in a different order than the data received from the server) or else look for an alternative solution.

It works the the bug that was reported about addons requirements.
In most cases, the text we hold in the backend cannot contain html (like cluster name, dns, state, etc).
So this attempts to resolve only the issue we see today regarding addon parameter requirements.
What I'm suggesting here is to resolve it for that use case only, and still be able to get the output sorted as it is in the backend.
WDYT?

jhernand:
This only works correctly when the output is a terminal. If you redirect to a file or use a pipe it won't work as you expect. If that is OK feel free to merge.

jhernand:
By the way, it would be nice to have a test for this. See https://github.com/openshift-online/ocm-cli/blob/main/tests/get_test.go for examples.

tzvatot:
> This only works correctly when the output is a terminal. If you redirect to a file or use a pipe it won't work as you expect. If that is OK feel free to merge.

QE approved this is sufficient. I will merge it.

PR #377, opened 2022-02-27
improving existing vpc and proxy code

zvikorn:

SDA-5487
SDA-5489
SDA-5491

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -892,6 +892,10 @@ func promptExistingGCPVPC(fs *pflag.FlagSet, connection *sdk.Connection) error {
 				args.existingVPC.ComputeSubnet)
 		}
 
+		if wasGCPNetworkReceived() {
+			args.existingVPC.Enabled = true
+		}
+
 		fs.Set("use-existing-vpc", "true")
 		flag := fs.Lookup("vpc-name")
 		if !flag.Changed {
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -399,7 +399,7 @@ func CreateCluster(cmv1Client *cmv1.Client, config Spec, dryRun bool) (*cmv1.Clu
 }
 
 func isGCPNetworkExists(existingVPC ExistingVPC) bool {
-	return existingVPC.VPCName != "" && existingVPC.ControlPlaneSubnet != "" && existingVPC.ComputeSubnet != ""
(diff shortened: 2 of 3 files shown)


cmd/ocm/create/cluster/cmd.go
```diff
@@ -579,7 +579,7 @@ func promptName(argv []string) error {
 
 func promptClusterWideProxy() error {
 	var err error
-	if args.existingVPC.Enabled && !wasClusterWideProxyReceived() && args.interactive {
+	if !args.existingVPC.Enabled && !wasClusterWideProxyReceived() && args.interactive {
```
tbrisker, line 582: Should this question be displayed regardless of if the existing vpc is enabled?

pkg/cluster/cluster.go
```diff
@@ -399,7 +399,7 @@ func CreateCluster(cmv1Client *cmv1.Client, config Spec, dryRun bool) (*cmv1.Clu
 }
 
 func isGCPNetworkExists(existingVPC ExistingVPC) bool {
-	return existingVPC.VPCName != "" && existingVPC.ControlPlaneSubnet != "" && existingVPC.ComputeSubnet != ""
+	return existingVPC.VPCName != "" || existingVPC.ControlPlaneSubnet != "" || existingVPC.ComputeSubnet != ""
```
tbrisker, line 402: Wouldn't this change cause the cluster builder to try building the cluster with the missing attributes and only fail then? 
Perhaps instead of using this method for [checking if we want to add these attributes](https://github.com/zvikorn/ocm-cli/blob/2e8877639fb3e90c90640346432e95bde7cc1bdc/pkg/cluster/cluster.go#L333) to the builder, we need to check if any of the attributes is missing and error early?

tbrisker (commented):
See inline questions. 
As a side note, it would have made the review easier if there was a separate PR for each issue. 1-line PRs are the best kind 😄

pkg/cluster/cluster.go
zvikorn: we build the GCPNetwork object if at least one GCPNetwork attribute exists. If CS needs all attributes it will return an error.

cmd/ocm/create/cluster/cmd.go
zvikorn: no. Only if vpc is enabled

tbrisker: so why is this check looking for `!args.existingVPC.Enabled`? that seems to mean the opposite, unless i'm missing something

zvikorn: yup. you right. Thanks!

PR #381, opened 2022-04-05
Correct the ocm json location in mac

datianshi:
As title


jhernand:
Looks good, thanks @datianshi.

PR #384, opened 2022-05-03
Validate labels and taints for MachinePool edit

bergmannf:
Prevent the user from passing invalid labels or taints to the machinepool and use the same validations kubernetes is using for these types.

While testing an alert for the https://github.com/openshift/managed-node-metadata-operator I realized it is possible to add invalid taints to the MachinePools.

This PR will run the validations before attempting to modify the MachinePool, so the errors are caught earlier.

For completeness I added the check to labels as well - even though the api-server correctly rejected invalid labels but not taints.

### DIFF
--- a/cmd/ocm/edit/machinepool/cmd.go
+++ b/cmd/ocm/edit/machinepool/cmd.go
@@ -22,6 +22,7 @@ import (
 	"github.com/openshift-online/ocm-cli/pkg/ocm"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/util/validation"
 )
 
 var args struct {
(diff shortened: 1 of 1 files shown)


cmd/ocm/edit/machinepool/cmd.go
```diff
@@ -125,6 +126,14 @@ func run(cmd *cobra.Command, argv []string) error {
 				return fmt.Errorf("Expected key=value format for label-match")
 			}
 			tokens := strings.Split(label, "=")
+			errors := validation.IsQualifiedName(tokens[0])
```
igoihman, line 129: The validation belongs in the server side and as far as I remember we already had it in place. Are you seeing errors when trying to create labels and taints with invalid names?

bergmannf: When trying to add a label that is invalid, I get an error (as I would expect):

```
ocm edit machinepool --cluster 1rvd68i76k3mkijnkhmm4ivlbpl4no5j --labels="invalid taint=invalid" mp-1
Error: Failed to edit machine pool for cluster '1rvd68i76k3mkijnkhmm4ivlbpl4no5j': status is 400, identifier is '400', code is 'CLUSTERS-MGMT-400' and operation identifier is '0bbdb158-c200-42af-a09c-a369bf5da580': Invalid label key 'invalid taint': name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')
```

When I attempt to add a taint that is invalid, `ocm` will allow creating an invalid taint:

```
ocm edit machinepool --cluster 1rvd68i76k3mkijnkhmm4ivlbpl4no5j --taints="invalid taint=invalid:PreferNoSchedule" mp-1
echo $?
0
```

In this case only the `effect` is verified to be one of `['NoExecute', 'NoSchedule', 'PreferNoSchedule']`, but the key and value are not validated.

PR #391, opened 2022-06-09
Added sub command limitedsupportreasons to List command

luis-falcon:
Added sub command limitedsupportreasons to List command in order to get the reasons why a cluster is in limited support.

Link to the Jira:
https://issues.redhat.com/browse/SDA-6109

### DIFF
--- a/cmd/ocm/list/limitedsupportreason/cmd.go
+++ b/cmd/ocm/list/limitedsupportreason/cmd.go
@@ -1,0 +1,151 @@
+/*
+Copyright (c) 2022 Red Hat, Inc.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+  http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package limitedsupportreason
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	c "github.com/openshift-online/ocm-cli/pkg/cluster"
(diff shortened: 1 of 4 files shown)


cmd/ocm/list/limitedsupportreason/cmd.go
```diff
@@ -0,0 +1,151 @@
+/*
+Copyright (c) 2020 Red Hat, Inc.
```
vkareh, line 2: This should be 2022

pkg/output/tables/limitedSupport.yaml
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (c) 2021 Red Hat, Inc.
```
vkareh, line 2: This should be 2022

luis-falcon: Done

cmd/ocm/list/limitedsupportreason/cmd.go
luis-falcon: Done

pkg/cluster/cluster.go
```diff
@@ -126,6 +126,11 @@ type AddOnItem struct {
 	Available bool
 }
 
+type lmtSprReasonItem struct {
+	ID      string
+	Summary string
+}
```
sam-nguyen7, line 132: Can we also include the `details` field as well?

sam-nguyen7 (changes requested):

pkg/cluster/cluster.go
luis-falcon: Not for now. maybe in another card for this, because then we have to mess with the table generation code or find a solution to output it differently (example create a details flag in order to change from a table format to list or find a way to make the table not wrap around). Issue is the table details column is long and overlaps a lot and makes the whole thing cluttered. 

Now, if we don't care about how pretty it looks i have no problem, we can add it. but then the UX for this command would be terrible.

sam-nguyen7 (approved):
/lgtm

PR #393, opened 2022-06-11
Correct the doc URLs for IDP creation

mrbarge:
I observed that the current interactive wizard for adding an IDP points to a documentation URL that doesn't redirect to IDP-specific documentation (it just lands at the OSD docs homepage).

This PR updates the links to point to the correct documentation resources.


igoihman (approved):
/lgtm

PR #395, opened 2022-07-03
better error when gcp ccs creds are not provided

zvikorn:

SDA-5492


zvikorn:
@zgalor @oriAdler @pvasant PTAL

oriAdler:
LGTM

PR #398, opened 2022-07-21
SDA-5813 - Remove DisplayName/display_name from cluster

hunterkepley:
https://issues.redhat.com/browse/SDA-5813

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -147,9 +147,9 @@ func GetCluster(connection *sdk.Connection, key string) (cluster *cmv1.Cluster, 
 
 	// Try to find a matching subscription:
 	subsSearch := fmt.Sprintf(
-		"(display_name = '%s' or cluster_id = '%s' or external_cluster_id = '%s') and "+
+		"(cluster_id = '%s' or external_cluster_id = '%s') and "+
 			"status in ('Reserved', 'Active')",
-		key, key, key,
+		key, key,
 	)
 	subsListResponse, err := subsResource.List().
 		Search(subsSearch).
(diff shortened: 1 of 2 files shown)


pvasant:
@zgalor PTAL

vkareh:
We still have `name` which is what customers name the cluster to begin with. Do we want to use that instead? In `rosa`, for example, we allow cluster selection by name: `$ rosa describe cluster -c mycluster-foo`

So maybe we want to use `name` instead of removing `display_name`

hunterkepley:
> We still have `name` which is what customers name the cluster to begin with. Do we want to use that instead? In `rosa`, for example, we allow cluster selection by name: `$ rosa describe cluster -c mycluster-foo`
> 
> So maybe we want to use `name` instead of removing `display_name`

Added `name = '%s'` to replace `display_name = '%s`.

Thanks for letting me know this!

hunterkepley:
Tested and `DisplayName` no longer shows when describing a cluster through ocm-cli

pkg/cluster/cluster.go
```diff
@@ -147,7 +147,7 @@ func GetCluster(connection *sdk.Connection, key string) (cluster *cmv1.Cluster,
 
 	// Try to find a matching subscription:
 	subsSearch := fmt.Sprintf(
-		"(display_name = '%s' or cluster_id = '%s' or external_cluster_id = '%s') and "+
+		"(name = '%s' or cluster_id = '%s' or external_cluster_id = '%s') and "+
```
zgalor, line 150: @hunterkepley this is an AMS field. Shouldn't change.
fixed in https://github.com/openshift-online/ocm-cli/pull/413

zgalor: cc @vkareh

PR #399, opened 2022-07-21
network: Ensure there is no default network type

vkareh:
Since the network type is defined in the API based on the OpenShift
version, we remove the client-side default. If the user overrides the
hidden flag, the API will ensure that the user has the required
capabilities to use the non-default network type for that version.


mrbarge:
This has been tested successfully. Thanks!
/lgtm

PR #400, opened 2022-08-01
Bump golang dependencies

cblecker:
This bumps all the direct dependencies, but specifically brings in `github.com/spf13/cobra v1.5.0`. This will fix some bash completion interaction issues with the backplane plugin.

Note that I only bumped `ocm-sdk-go` to v0.1.275 instead of v0.1.278.. we'd need a new release of the rosa dep to go higher, because there is a breaking change.


pvasant:
why not 0.1.277 ?  0.1.278 has display name change.

cblecker:
@pvasant I can bump higher if you like. I just matched the version from the current ROSA release. Would you like me to bump to 277?

pvasant:
nahh i m good to merge. Just was not sure.  I think @hunterkepley is anyways working on modifying clis  to latest SDK version

PR #404, opened 2022-08-23
Add no-proxy attribute to OCM-CLI

zvikorn:

SDA-5613

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -383,6 +383,12 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		return err
 	}
 
+	if wasClusterWideProxyReceived() {
+		args.ccs.Enabled = true
+		args.existingVPC.Enabled = true
+		args.clusterWideProxy.Enabled = true
+	}
+
 	err = promptCCS(fs)
 	if err != nil {
 		return err
@@ -462,14 +468,6 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		if err != nil {
 			return err
 		}
-	}
-
-	args.existingVPC.Enabled = false
-	args.clusterWideProxy.Enabled = false
-	if wasClusterWideProxyReceived() {
-		args.ccs.Enabled = true
-		args.existingVPC.Enabled = true
-		args.clusterWideProxy.Enabled = true
 	}
 
 	if args.existingVPC.SubnetIDs != "" {
@@ -613,9 +611,7 @@ func promptClusterWideProxy() error {
 				}
 				args.existingVPC.Enabled = true
 			}
-		}
 
-		if args.clusterWideProxy.Enabled {
 			if args.clusterWideProxy.HTTPSProxy == nil {
 				args.clusterWideProxy.HTTPSProxy = new(string)
 			}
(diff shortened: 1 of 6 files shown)


cmd/ocm/create/cluster/cmd.go
```diff
@@ -78,6 +79,14 @@ const clusterNameHelp = "will be used when generating a sub-domain for your clus
 
 const subnetTemplate = "%s (%s)"
 
+// the following regex defines five different patterns:
```
tbrisker, line 82: There's only four patterns here

cmd/ocm/create/cluster/cmd.go
```diff
@@ -78,6 +79,14 @@ const clusterNameHelp = "will be used when generating a sub-domain for your clus
...
+// first pattern is to validate IPv4 address
+// second,is for IPv4 CIDR range validation
+// third pattern is to validate domains
+// and the forth petterrn is to be able to remove the existing no-proxy value by typing empty string ("").
```
tbrisker, line 86: np, spelling:
```suggestion
// and the fourth pattern is to be able to remove the existing no-proxy value by typing empty string ("").
```

cmd/ocm/create/cluster/cmd.go
```diff
@@ -78,6 +79,14 @@ const clusterNameHelp = "will be used when generating a sub-domain for your clus
...
+// nolint
+var UserNoProxyRE = regexp.MustCompile(`^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$|^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/(3[0-2]|[1-2][0-9]|[0-9]))$|^(.?[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9][a-z0-9-]{0,61}[a-z0-9]$|^""$`)
```
tbrisker, line 88: Yikes!
Instead of this monstrous incantation, could we instead use something like:
```go
func isValidNoProxy(val string) bool {
    return val==`""` || isIp(val) || isDomain(val) || isCIDR(val)
}
```
?

cmd/ocm/create/cluster/cmd.go
```diff
@@ -383,6 +392,14 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		return err
 	}
 
+	args.existingVPC.Enabled = false
+	args.clusterWideProxy.Enabled = false
```
tbrisker, line 396: NP: no need to explicitly set these to `false`, that is the zero value of bool

cmd/ocm/create/cluster/cmd.go
```diff
@@ -613,9 +622,7 @@ func promptClusterWideProxy() error {
 				}
 				args.existingVPC.Enabled = true
 			}
-		}
 
-		if args.clusterWideProxy.Enabled {
```
tbrisker, line 618: why are we removing the check here?

cmd/ocm/edit/cluster/cmd.go
```diff
@@ -179,29 +188,30 @@ func run(cmd *cobra.Command, argv []string) error {
...
+				return fmt.Errorf("Invalid 'proxy.https_proxy' attribute '%s'", *args.clusterWideProxy.HTTPSProxy)
 			}
 		}
-		httpsProxy = &httpsProxyValue
+		httpsProxy = args.clusterWideProxy.HTTPSProxy
+	}
+
+	var noProxy *string
+	if cmd.Flags().Changed("no-proxy") {
+		noProxy = args.clusterWideProxy.NoProxy
```
tbrisker, line 214: Should there be some validation here?

pkg/cluster/cluster.go
```diff
@@ -450,6 +457,16 @@ func UpdateCluster(client *cmv1.ClustersClient, clusterID string, config Spec) e
 		}
 		clusterBuilder = clusterBuilder.Proxy(clusterProxyBuilder)
 	}
+	if config.ClusterWideProxy.NoProxy != nil {
+		if (config.ClusterWideProxy.HTTPProxy != nil && *config.ClusterWideProxy.HTTPProxy != "") ||
+			(config.ClusterWideProxy.HTTPSProxy != nil && *config.ClusterWideProxy.HTTPSProxy != "") ||
+			cluster.Body().Proxy().HTTPProxy() != "" || cluster.Body().Proxy().HTTPSProxy() != "" {
```
tbrisker, line 463: I wonder if we should skip this check here and leave it to the backend to validate there is a proxy when setting noproxy, that way we don't have to fetch the cluster first and can unify adding the NoProxy section with the previous condition

cmd/ocm/create/cluster/cmd.go
zvikorn: This is how we defined it in CS as well as in ROSA. If we want to change that we may open a new ticket for that. This will require a comprehensive regression test too.

cmd/ocm/create/cluster/cmd.go
zvikorn: correct. Thanks!

cmd/ocm/create/cluster/cmd.go
zvikorn: we just moving it.

pkg/cluster/cluster.go
zvikorn: Most of the code that I was involved with in both in ocm-cli and rosa, contains as much validation as possible to avoid sending the request to CS and let it decide instead of the client.

cmd/ocm/edit/cluster/cmd.go
zvikorn: yes. Thanks!

pkg/cluster/cluster.go
tbrisker: but to make this validation work we need to send a request to CS to get the cluster first, so we aren't saving a request here - in fact we're adding a request to all update calls, even if we don't touch the proxy

cmd/ocm/create/cluster/cmd.go
zvikorn: https://github.com/openshift-online/ocm-cli/pull/404/files#diff-2f19822c49dc0bec88a73920ea0f04a30b7047b5a827c90447e02b7378744f97R593
There was another one existed already

PR #407, opened 2022-08-30
Add validation when user creates a cluster only with no-proxy (non-interacitve mode)

zvikorn:

SDA-6712

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -688,6 +688,11 @@ func promptClusterWideProxy() error {
 			}
 		}
 	}
+	if (args.clusterWideProxy.HTTPProxy != nil && *args.clusterWideProxy.HTTPProxy == "") &&
+		(args.clusterWideProxy.HTTPSProxy != nil && *args.clusterWideProxy.HTTPSProxy == "") &&
+		(args.clusterWideProxy.NoProxy != nil && *args.clusterWideProxy.NoProxy != "") {
+		return fmt.Errorf("Expected at least one of the following: http-proxy, https-proxy")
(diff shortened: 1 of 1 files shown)


cmd/ocm/create/cluster/cmd.go
```diff
@@ -688,6 +688,11 @@ func promptClusterWideProxy() error {
 			}
 		}
 	}
+	if (args.clusterWideProxy.HTTPProxy != nil && *args.clusterWideProxy.HTTPProxy == "") &&
+		(args.clusterWideProxy.HTTPSProxy != nil && *args.clusterWideProxy.HTTPSProxy == "") &&
+		*args.clusterWideProxy.NoProxy != "" {
```
zvikorn, line 693: args.clusterWideProxy.NoProxy cannot be nil

tbrisker: are you sure?
https://github.com/openshift-online/ocm-cli/pull/407/files#diff-2f19822c49dc0bec88a73920ea0f04a30b7047b5a827c90447e02b7378744f97R648

zvikorn: https://github.com/openshift-online/ocm-cli/blob/main/pkg/arguments/arguments.go#L224

tbrisker: The link didn't work as expected, but take a look at line 648 where it's being set to `nil`

zvikorn: Weird... I have tested all options I was thinking about. I assume when we set it with "", the length is not 0?
Anyway, I am adding the check for nil

PR #413, opened 2022-11-14
Fix GetCluster AMS search

zgalor:
Related: SDA-7270


zgalor:
@cben @oriAdler @tzvatot PTAL

tzvatot:
LGTM

PR #416, opened 2022-12-16
Adds Management Cluster field for Hypershift Clusters describe

fahlmant:
Adds the `Management Cluster` field to `ocm describe cluster` for Hypershift Hosted Control Plane clusters. This is the equivalent of `Shard` for classic OSD/ROSA.


pkg/cluster/describe.go
```diff
@@ -157,6 +157,21 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
...
+	mgmtCluster := ""
+	if cluster.Hypershift().Enabled() {
+		hypershiftResp, err := connection.ClustersMgmt().V1().Clusters().
+			Cluster(cluster.ID()).
+			Hypershift().
+			Get().
+			Send()
+		if err != nil {
+			fmt.Println(err)
+			return err
```
mjlshen, line 170:
```suggestion
				return fmt.Errorf(
					"can't get management cluster for '%s': %v",
					cluster.ID(), err,
				)
```

mjlshen (changes requested):
We should return errors up instead of printing them

mjlshen (commented):
/lgtm

rafael-azevedo (commented):
/lgtm !

pkg/cluster/describe.go
```diff
@@ -157,6 +157,20 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
...
+	// Check Hypershift-related values
+	mgmtCluster := ""
+	if cluster.Hypershift().Enabled() {
+		hypershiftResp, err := connection.ClustersMgmt().V1().Clusters().
+			Cluster(cluster.ID()).
+			Hypershift().
+			Get().
+			Send()
+		if err != nil {
+			return fmt.Errorf("unable to query hypershift endpoint for cluster %s: %v", cluster.ID(), err)
```
vkareh, line 169: This error should be silently ignored as there are only a handful of roles that can access this endpoint.

vkareh (approved):
/lgtm

PR #430, opened 2023-01-16
Remove validation for GCP+private clusters

cblecker:
This client-side validation is no longer needed. GCP + Private Clusters is supported, and specific options should be validated server-side.


igoihman (approved):
/lgtm

PR #441, opened 2023-02-10
Return the service cluster for HyperShift clusters with `ocm describe cluster`

mjlshen:
When troubleshooting HyperShift clusters, sometimes it is important to inspect the relevant service cluster. This PR does this by essentially implementing:

```
❯ ocm get /api/osd_fleet_mgmt/v1/service_clusters -p search="id='$(ocm get /api/osd_fleet_mgmt/v1/management_clusters -p search="name='hs-mc-op846kea0'" | jq -r '.items[0].parent.id')'" | jq -r '.items[0].name'

hs-sc-op846kde0
```

Sample output:
```
❯ ocm describe cluster 21rrnvvf6rd4vdndl4lf3gie7mat0gna                                

ID:			21rrnvvf6rd4vdndl4lf3gie7mat0gna
External ID:		d61a016b-dd48-4f96-ac27-dc9a1b1f7b63
...
Management Cluster:     hs-mc-jsui4ri9g
Service Cluster:        hs-sc-op846kde0

```

### DIFF
--- a/pkg/cluster/describe.go
+++ b/pkg/cluster/describe.go
@@ -157,17 +157,8 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster) 
 		isExistingVPC = "true"
 	}
 
-	// Check Hypershift-related values
-	mgmtCluster := ""
-	if cluster.Hypershift().Enabled() {
-		// Ignorning the error here as this endpoint is behind a specific permissioon
-		hypershiftResp, _ := connection.ClustersMgmt().V1().Clusters().
-			Cluster(cluster.ID()).
-			Hypershift().
-			Get().
-			Send()
-		mgmtCluster = hypershiftResp.Body().ManagementCluster()
-	}
+	// Parse Hypershift-related values
+	mgmtClusterName, svcClusterName := findHyperShiftMgmtSvcClusters(connection, cluster)
 
 	// Print short cluster description:
 	fmt.Printf("\n"+
@@ -235,26 +226,35 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster) 
 		cluster.CreationTimestamp().Round(time.Second).Format(time.RFC3339Nano),
 		cluster.ExpirationTimestamp().Round(time.Second).Format(time.RFC3339Nano),
 	)
+
+	// Hive
 	if shard != "" {
 		fmt.Printf("Shard:			%v\n", shard)
 	}
-	// This should be mutually exclusive with shard as it's hypershift specific
-	if mgmtCluster != "" {
-		fmt.Printf("Management Cluster:     %s\n", mgmtCluster)
+
+	// HyperShift (should be mutually exclusive with Hive)
+	if mgmtClusterName != "" {
+		fmt.Printf("Management Cluster:     %s\n", mgmtClusterName)
 	}
+	if svcClusterName != "" {
+		fmt.Printf("Service Cluster:        %s\n", svcClusterName)
+	}
+
+	// Cluster-wide-proxy
 	if cluster.Proxy().HTTPProxy() != "" {
 		fmt.Printf("HTTPProxy:	        %s\n", cluster.Proxy().HTTPProxy())
 	}
 	if cluster.Proxy().HTTPSProxy() != "" {
 		fmt.Printf("HTTPSProxy:	        %s\n", cluster.Proxy().HTTPSProxy())
 	}
-
 	if cluster.Proxy().NoProxy() != "" {
 		fmt.Printf("NoProxy:	        %s\n", cluster.Proxy().NoProxy())
 	}
 	if cluster.AdditionalTrustBundle() != "" {
 		fmt.Printf("AdditionalTrustBundle:  %s\n", cluster.AdditionalTrustBundle())
 	}
+
+	// GCP
 	if cluster.GCPNetwork().VPCName() != "" {
 		fmt.Printf("VPC-Name:	        %s\n", cluster.GCPNetwork().VPCName())
 	}
@@ -264,6 +264,8 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster) 
 	if cluster.GCPNetwork().ComputeSubnet() != "" {
 		fmt.Printf("Compute-Subnet:	        %s\n", cluster.GCPNetwork().ComputeSubnet())
 	}
+
+	// Limited Support Status
 	if cluster.Status().LimitedSupportReasonCount() > 0 {
 		fmt.Printf("Limited Support:	%t\n", cluster.Status().LimitedSupportReasonCount() > 0)
 	}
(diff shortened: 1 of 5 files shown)


pkg/cluster/describe.go
```diff
@@ -158,15 +159,30 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
...
+
+		mgmtClusterId = fmMgmtResp.Items().Get(0).ClusterManagementReference().ClusterId()
+
+		// href is in the format /api/osd_fleet_mgmt/v1/service_clusters/${id}
+		// and we are using it because ocm-sdk-go currently does not export the id directly
+		if kind := fmMgmtResp.Items().Get(0).Parent().Kind(); kind == "ServiceCluster" {
+			href := fmMgmtResp.Items().Get(0).Parent().Href()
+			_, id := path.Split(href)
+			fmSvcResp, _ := connection.OSDFleetMgmt().V1().ServiceClusters().ServiceCluster(id).Get().Send()
+			svcClusterId = fmSvcResp.Body().ClusterManagementReference().ClusterId()
```
mjlshen, line 184: The main reason I switched the management cluster to show its internal ID instead of the name as it was previously was due to the cluster name of the service cluster not being available here - I'd have to make an additional call to ClustersMgmt in order to get the name of a service cluster.

Happy to do so if that's preferred, but I felt that it was ok to show the ID instead since those are guaranteed to be unique.

cc @fahlmant

fahlmant: Either way works, this is for the SRE mostly, so its up to us to decide. We're already making multiple API calls, so I don't think one more would hurt, and reasoning about a name like `hs-mc-xxxxx` is a bit easier than an ID, but we use both constantly and internal ID is unique.

tl;dr either way works, there's an argument for both ways being a better UX so I have no idea what we should do

mjlshen: Got it - yeah agreed that `hs-mc-xxxx`/`hs-sc-xxxx` is easier to grok if this is mostly for SRE. I'll make that change

gdbranco: Is there anyway to enforce this only shows up for specific roles/permissions taken from token? Don't think usual clients will get much from this info

fahlmant: During printing, we already do this check 
```
if svcClusterId != "" {
```

mjlshen: also https://github.com/openshift-online/ocm-cli/pull/441/files#diff-2f84bc16b5d89dac13d11d6f8ae66c242bc75cccb194fdd0ace7723bb75b9312R165 we ignore errors and don't print anything by default - so it will only print if the specific roles/permissions are in place

pkg/cluster/describe_test.go
```diff
@@ -0,0 +1,40 @@
...
+	tests := []struct {
+		name         string
+		cluster      *cmv1.Cluster
+		expectedMgmt string
+		expectedSvc  string
+	}{
+		{
+			name:    "Not HyperShift",
+			cluster: newTestCluster(t, cmv1.NewCluster().Hypershift(cmv1.NewHypershift().Enabled(false))),
+		},
```
mjlshen, line 28: The current test-case covers the really simple case where this is not a HyperShift cluster and I'm not sure how to mock `conn *sdk.Connection` so that I can unit test HyperShift clusters. If there are any tips or examples I am happy to do so

pkg/cluster/describe.go
```diff
@@ -272,3 +275,48 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
...
+		href := fmMgmtResp.Items().Get(0).Parent().Href()
+		_, id := path.Split(href)
+		fmSvcResp, _ := conn.OSDFleetMgmt().V1().ServiceClusters().ServiceCluster(id).Get().Send()
+		svcClusterId := fmSvcResp.Body().ClusterManagementReference().ClusterId()
+
+		svcClusterResp, err := conn.ClustersMgmt().V1().Clusters().Cluster(svcClusterId).Get().Send()
+		if err != nil {
+			return mgmtClusterName, ""
+		}
+		return mgmtClusterName, svcClusterResp.Body().Name()
```
vkareh, line 317: Why are you looking up the cluster ID if you're returning the name? That should be available in the MC payload already:
```
if kind := fmMgmtResp.Items().Get(0).Parent().Kind(); kind == "ServiceCluster" {
  svcClusterName := fmMgmtResp.Items().Get(0).Parent().Name()
}
```

vkareh: okay, just saw the the SDK doesn't expose the SC name, but it's part of the payload. Can you add this to the corresponding model file? https://github.com/openshift-online/ocm-api-model/blob/main/model/osd_fleet_mgmt/v1/management_cluster_parent_type.model

mjlshen: Name isn't exported by ocm-sdk-go :( It's returned when I hit the API via `ocm get` though

pkg/cluster/describe.go
```diff
@@ -272,3 +275,48 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
...
+		Get().
+		Send()
+	if err != nil {
+		return "", ""
+	}
+
+	mgmtClusterName := hypershiftResp.Body().ManagementCluster()
+	fmMgmtResp, err := conn.OSDFleetMgmt().V1().ManagementClusters().
+		List().
+		Parameter("name", mgmtClusterName).
```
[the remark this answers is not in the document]
mjlshen: This does work, maybe via undefined behavior though :D - happy to change to this if that's how it's supposed to work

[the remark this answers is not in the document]
tiwillia: I'm also very surprised this works... what the heck

pkg/cluster/describe.go
mjlshen: Added to the model file in https://github.com/openshift-online/ocm-api-model/pull/690 and verified it worked - but now this PR should not be merged until that one is + a corresponding bump in ocm-sdk-go + a dependency bump in ocm-cli. So I am marking this PR as WIP.

vkareh: There's now a new SDK version with support for service/management cluster names. Can you update this?

mjlshen: Done! Thanks for pushing the SDK update forward!

vkareh:
@mjlshen do you still consider this `WIP`?

mjlshen:
Nope, thanks for the reminder - I've removed `WIP`

pkg/cluster/describe.go
```diff
@@ -272,3 +274,37 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
...
+	hypershiftResp, err := conn.ClustersMgmt().V1().Clusters().
+		Cluster(cluster.ID()).
+		Hypershift().
+		Get().
+		Send()
+	if err != nil {
+		return "", ""
+	}
+
+	mgmtClusterName := hypershiftResp.Body().ManagementCluster()
```
igoihman, line 295: what will be the output of this command for users who can't access the management cluster details?

Issue #517, opened 2023-07-04
`ocm describe cluster` fails for Archived/Deprovisioned clusters


cben:
here is an example Deprovisioned subscription
```json
> ocm get sub 2QaydcSjzvQGWHqxRpSReT7mWW4
{
  ...
  "cluster_id": "24301dug094uiivh57o0hfdhotp1cbom",
  "display_name": "celia-imdsv2",
  "external_cluster_id": "89ffcb09-33bd-47ce-aa21-e793383b136d",
  "href": "/api/accounts_mgmt/v1/subscriptions/2QaydcSjzvQGWHqxRpSReT7mWW4",
  "id": "2QaydcSjzvQGWHqxRpSReT7mWW4",
  "kind": "Subscription",
  "managed": true,
  ...
  "organization_id": "1wuVGGV6SCmD8ya6yRGEJzvmVuC",
  "plan": {
    "href": "/api/accounts_mgmt/v1/plans/MOA",
    ...
  },
  "provenance": "Provisioning",
  "status": "Deprovisioned",
}
```

and now let's try to `ocm describe cluster` it:
```
> ocm describe cluster 89ffcb09-33bd-47ce-aa21-e793383b136d
Error: Can't retrieve cluster for key '89ffcb09-33bd-47ce-aa21-e793383b136d': There are no subscriptions or clusters with identifier or name '89ffcb09-33bd-47ce-aa21-e793383b136d'

> ocm describe cluster celia-imdsv2
Error: Can't retrieve cluster for key 'celia-imdsv2': There are no subscriptions or clusters with identifier or name 'celia-imdsv2'

> ocm describe cluster 2QaydcSjzvQGWHqxRpSReT7mWW4
Error: Can't retrieve cluster for key '2QaydcSjzvQGWHqxRpSReT7mWW4': There are no subscriptions or clusters with identifier or name '2QaydcSjzvQGWHqxRpSReT7mWW4'

> ocm describe cluster 24301dug094uiivh57o0hfdhotp1cbom
Error: Can't retrieve cluster for key '24301dug094uiivh57o0hfdhotp1cbom': There are no subscriptions or clusters with identifier or name '24301dug094uiivh57o0hfdhotp1cbom'
```

These messages are both misleading technically (there _is_ a subscription with these ids) and unfortunate :frowning_face:.

Here are the accessed APIs:
```
[...]

PR #518, opened 2023-07-04
improve GetCluster message when Subscription exists but is inactive

cben:
First step towards #517.
At least doesn't lie to user and gives hint how to find past data:

#### before
```
> ocm describe cluster 0e68a621-c00e-4bbf-a4bb-9915ae6cc947
Error: Can't retrieve cluster for key '0e68a621-c00e-4bbf-a4bb-9915ae6cc947': There are no subscriptions or clusters with identifier or name '0e68a621-c00e-4bbf-a4bb-9915ae6cc947'
```
#### after
```
> ocm describe cluster 0e68a621-c00e-4bbf-a4bb-9915ae6cc947
Error: Can't retrieve cluster for key '0e68a621-c00e-4bbf-a4bb-9915ae6cc947': Cluster was Deprovisioned, see `ocm get subscription 2S434jffVG4fSgDdK4cbkDTxVNp` for details``
```

This affects ALL commands that use `GetCluster` helper.

:question: Behavior change: when there is 1 active cluster and also 1+ past clusters reusing same display_name, previously the active one was silently used; now will refuse treating them as ambiguous:
```
> ocm get subs -p search="display_name in ('servheredia')" | jq .items[].status | sort | uniq -c
      4 "Deprovisioned"
      1 "Reserved"
> ocm describe cluster servheredia
Error: Can't retrieve cluster for key 'servheredia': There are 5 subscriptions with cluster identifier or name 'servheredia'
```
is that desired?

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -148,8 +148,7 @@ func GetCluster(connection *sdk.Connection, key string) (cluster *cmv1.Cluster, 
 
 	// Try to find a matching subscription:
 	subsSearch := fmt.Sprintf(
-		"(display_name = '%s' or cluster_id = '%s' or external_cluster_id = '%s') and "+
-			"status in ('Reserved', 'Active')",
+		"(display_name = '%s' or cluster_id = '%s' or external_cluster_id = '%s')",
 		key, key, key,
 	)
 	subsListResponse, err := subsResource.List().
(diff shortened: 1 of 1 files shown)


gdbranco (approved):

mjlshen:
I'm curious about this change:
> ❓ Behavior change: when there is 1 active cluster and also 1+ past clusters reusing same display_name, previously the active one was silently used; now will refuse treating them as ambiguous:

The behavior of `ocm describe cluster` to just show the active cluster was desired by me. When I spin up a test cluster I tend to keep the same name, so now when I want to us `ocm describe cluster` I would have to use one of the IDs instead which are unpredictable.

I appreciate that we want a way to query archived clusters with ocm cli - but is there a flag we can add to keep the old behavior, maybe `--active`? Alternatively, the opposite and keeping the old behavior by default and adding a `--archives`?

cben:
> When I spin up a test cluster I tend to keep the same name

Good point.  I'm less involved with ocm-cli now so I'm not the right person to decide this. I recommend opening a new issue.

cben:
Or bring this up on #517.  One of the open questions there was that many commands using GetCluster() only make sense on Active clusters...  
(This PR was not so much a solution as provocation to explore the question 😉)

PR #531, opened 2023-07-31
OCM-2966 : Feat : Added subscription_type parameter to create cluster command

tirthct:
**Changed**
- Added subscription_type parameter to create cluster command
- Added validation to have one of the specific options for the command
- Defaulted subscription_type to standard

**Tested**
- Cluster should be created based on the subscription_type parameter passed in cli : tested

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -39,11 +39,16 @@ import (
 	"github.com/spf13/pflag"
 )
 
+var ValidSubscriptionTypes = []string{standardSubscriptionType, marketplaceRhmSubscriptionType, marketplaceGcpSubscriptionType}
+
 const (
 	defaultIngressRouteSelectorFlag            = "default-ingress-route-selector"
 	defaultIngressExcludedNamespacesFlag       = "default-ingress-excluded-namespaces"
 	defaultIngressWildcardPolicyFlag           = "default-ingress-wildcard-policy"
 	defaultIngressNamespaceOwnershipPolicyFlag = "default-ingress-namespace-ownership-policy"
+	standardSubscriptionType                   = "standard"
+	marketplaceRhmSubscriptionType             = "marketplace-rhm"
+	marketplaceGcpSubscriptionType             = "marketplace-gcp"
 )
 
 var args struct {
@@ -68,6 +73,7 @@ var args struct {
 	clusterWideProxy      c.ClusterWideProxy
 	gcpServiceAccountFile arguments.FilePath
 	etcdEncryption        bool
+	subscriptionType      string
 
 	// Scaling options
 	computeMachineType string
@@ -301,6 +307,14 @@ func init() {
 		fmt.Sprintf("Namespace Ownership Policy for ingress. Options are %s",
 			strings.Join(ingress.ValidNamespaceOwnershipPolicies, ",")),
 	)
+
+	fs.StringVar(
+		&args.subscriptionType,
+		"subscription-type",
+		standardSubscriptionType,
+		fmt.Sprintf("The subscription billing model for the cluster. Options are %s",
+			strings.Join(ValidSubscriptionTypes, ",")),
+	)
 }
 
 func osdProviderOptions(_ *sdk.Connection) ([]arguments.Option, error) {
@@ -375,6 +389,18 @@ func getVersionOptionsWithDefault(connection *sdk.Connection, channelGroup strin
 		})
 	}
 	return
+}
+
+func getSubscriptionTypeOptions() []arguments.Option {
+	options := []arguments.Option{}
+	for _, subscriptionType := range ValidSubscriptionTypes {
+		description := ""
+		options = append(options, arguments.Option{
+			Value:       subscriptionType,
+			Description: description,
+		})
+	}
+	return options
 }
 
 func getMachineTypeOptions(connection *sdk.Connection) ([]arguments.Option, error) {
@@ -530,6 +556,11 @@ func preRun(cmd *cobra.Command, argv []string) error {
 	if err != nil {
 		return err
 	}
+
+	if err = arguments.CheckOneOf(fs, "subscription-type", getSubscriptionTypeOptions()); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -577,6 +608,7 @@ func run(cmd *cobra.Command, argv []string) error {
 		Private:            &args.private,
 		EtcdEncryption:     args.etcdEncryption,
 		DefaultIngress:     defaultIngress,
+		SubscriptionType:   args.subscriptionType,
 	}
 
 	cluster, err := c.CreateCluster(connection.ClustersMgmt().V1(), clusterConfig, args.dryRun)
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -94,6 +94,8 @@ const clusterNameHelp = "will be used when generating a sub-domain for your clus
 
 const subnetTemplate = "%s (%s)"
 
+const subscriptionTypeTemplate = "%s (%s)"
+
 // Creates a subnet options using a predefined template.
 func setSubnetOption(subnet, zone string) string {
 	return fmt.Sprintf(subnetTemplate, subnet, zone)
@@ -102,6 +104,14 @@ func setSubnetOption(subnet, zone string) string {
 // Parses the subnet from the option chosen by the user.
 func parseSubnet(subnetOption string) string {
 	return strings.Split(subnetOption, " ")[0]
+}
+
+func setSubscriptionTypeOption(id, description string) string {
+	return fmt.Sprintf(subscriptionTypeTemplate, id, description)
+}
+
+func parseSubscriptionType(subscriptionTypeOption string) string {
+	return strings.Split(subscriptionTypeOption, " ")[0]
 }
 
 // Cmd Constant:
@@ -396,7 +406,7 @@ func getSubscriptionTypeOptions(connection *sdk.Connection) ([]arguments.Option,
 		return options, err
 	}
 	for _, billingModel := range billingModels {
-		option := subscriptionTypeTemplate(billingModel.Id(), billingModel.Description())
+		option := subscriptionTypeOption(billingModel.Id(), billingModel.Description())
 		//Standard billing model should always be the first option
 		if billingModel.Id() == billing.StandardSubscriptionType {
 			if len(options) == 0 { // nil or empty slice or after last element
@@ -412,21 +422,11 @@ func getSubscriptionTypeOptions(connection *sdk.Connection) ([]arguments.Option,
 	return options, nil
 }
 
-func subscriptionTypeTemplate(id string, description string) arguments.Option {
+func subscriptionTypeOption(id string, description string) arguments.Option {
 	option := arguments.Option{
-		Value: fmt.Sprintf("%s (%s)", id, description),
+		Value: setSubscriptionTypeOption(id, description),
 	}
 	return option
-}
-
-func getSubscriptionTypeIdFromDescription(connection *sdk.Connection, value string) string {
-	billingModels, _ := billing.GetBillingModels(connection)
-	for _, billingModel := range billingModels {
-		if value == fmt.Sprintf("%s (%s)", billingModel.Id(), billingModel.Description()) {
-			return billingModel.Id()
-		}
-	}
-	return ""
 }
 
 func getMachineTypeOptions(connection *sdk.Connection) ([]arguments.Option, error) {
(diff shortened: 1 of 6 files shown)


cmd/ocm/create/cluster/cmd.go
```diff
@@ -39,11 +39,16 @@ import (
...
+var ValidSubscriptionTypes = []string{standardSubscriptionType, marketplaceRhmSubscriptionType, marketplaceGcpSubscriptionType}
+
 const (
 	defaultIngressRouteSelectorFlag            = "default-ingress-route-selector"
 	defaultIngressExcludedNamespacesFlag       = "default-ingress-excluded-namespaces"
 	defaultIngressWildcardPolicyFlag           = "default-ingress-wildcard-policy"
 	defaultIngressNamespaceOwnershipPolicyFlag = "default-ingress-namespace-ownership-policy"
+	standardSubscriptionType                   = "standard"
+	marketplaceRhmSubscriptionType             = "marketplace-rhm"
+	marketplaceGcpSubscriptionType             = "marketplace-gcp"
```
cristianoveiga, line 51: This will map to existing billing models. 

The possible values are listed here: https://github.com/openshift-online/ocm-api-model/blob/main/model/accounts_mgmt/v1/billing_model_type.model#L18-L40

cristianoveiga:
Let's update the PR name/commit description with the latest convention

cmd/ocm/create/cluster/cmd.go
```diff
@@ -377,6 +393,56 @@ func getVersionOptionsWithDefault(connection *sdk.Connection, channelGroup strin
...
 
+func getSubscriptionTypeOptions(connection *sdk.Connection) ([]arguments.Option, error) {
+	options := []arguments.Option{}
+	billingModels, err := getBillingModels(connection)
+	if err != nil {
+		return options, err
+	}
+	for _, billingModel := range billingModels {
+		options = append(options, arguments.Option{
+			Value:       fmt.Sprintf("%s (%s)", billingModel.Id(), billingModel.Description()),
```
cristianoveiga, line 404: Why do we need to add the description to the value?

tirthct: Added it before since interactive mode has a ? option to show description, but seems it only shows descriptions of the argument, not the drop down. Removed the description

cmd/ocm/create/cluster/cmd.go
```diff
@@ -301,6 +307,14 @@ func init() {
 		fmt.Sprintf("Namespace Ownership Policy for ingress. Options are %s",
 			strings.Join(ingress.ValidNamespaceOwnershipPolicies, ",")),
 	)
+
+	fs.StringVar(
+		&args.subscriptionType,
+		"subscription-type",
+		standardSubscriptionType,
+		fmt.Sprintf("The subscription billing model for the cluster. Options are %s",
+			strings.Join(ValidSubscriptionTypes, ",")),
```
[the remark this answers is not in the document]
tirthct: Done

cmd/ocm/create/cluster/cmd.go
```diff
@@ -377,6 +394,68 @@ func getVersionOptionsWithDefault(connection *sdk.Connection, channelGroup strin
...
+		} else {
+			options = append(options, arguments.Option{
+				Value: fmt.Sprintf("%s (%s)", billingModel.Id(), billingModel.Description()),
+			})
+		}
+	}
+	return options, nil
+}
+
+func getSubscriptionTypeIdFromDescription(connection *sdk.Connection, value string) string {
```
cristianoveiga, line 424: I would suggest using the same approach used for subnets, see https://github.com/openshift-online/ocm-cli/blob/main/cmd/ocm/create/cluster/cmd.go#L885

cmd/ocm/create/cluster/cmd.go
```diff
@@ -377,6 +394,68 @@ func getVersionOptionsWithDefault(connection *sdk.Connection, channelGroup strin
...
+	billingModels, _ := getBillingModels(connection)
+	for _, billingModel := range billingModels {
+		if value == fmt.Sprintf("%s (%s)", billingModel.Id(), billingModel.Description()) {
+			return billingModel.Id()
+		}
+	}
+	return ""
+}
+
+func getBillingModel(connection *sdk.Connection, billingModelID string) (*amv1.BillingModelItem, error) {
```
cristianoveiga, line 434: We can make this function reusable (as well as getBillingModels) by moving them to its own package under pkg/billing, for example.

This logic doesn't really belong to cmd/ocm/create/cluster/cmd.go

cmd/ocm/create/cluster/cmd.go
```diff
@@ -377,6 +394,68 @@ func getVersionOptionsWithDefault(connection *sdk.Connection, channelGroup strin
...
+	options := []arguments.Option{}
+	billingModels, err := getBillingModels(connection)
+	if err != nil {
+		return options, err
+	}
+	for _, billingModel := range billingModels {
+		//Standard billing model should always be the first option
+		if billingModel.Id() == StandardSubscriptionType {
+			standardOption := arguments.Option{
+				Value: fmt.Sprintf("%s (%s)", billingModel.Id(), billingModel.Description()),
```
cristianoveiga, line 407: This logic is duplicated and probably deserves its own function

cmd/ocm/create/cluster/cmd.go
```diff
@@ -377,6 +394,68 @@ func getVersionOptionsWithDefault(connection *sdk.Connection, channelGroup strin
...
 }
 
+func getSubscriptionTypeOptions(connection *sdk.Connection) ([]arguments.Option, error) {
+	options := []arguments.Option{}
+	billingModels, err := getBillingModels(connection)
+	if err != nil {
+		return options, err
+	}
+	for _, billingModel := range billingModels {
+		//Standard billing model should always be the first option
```
cristianoveiga, line 404: Can we refactor/simplify this?

cmd/ocm/create/cluster/cmd.go
```diff
@@ -421,12 +500,28 @@ func preRun(cmd *cobra.Command, argv []string) error {
...
+
 	// Only offer the 2 providers known to support OSD now;
 	// but don't validate if set, to not block `ocm` CLI from creating clusters on future providers.
 	providers, _ := osdProviderOptions(connection)
-	err = arguments.PromptOneOf(fs, "provider", providers)
-	if err != nil {
-		return err
+	// If marketplace-gcp subscription type is used, provider can only be GCP
+	gcpBillingModel, _ := getBillingModel(connection, marketplaceGcpSubscriptionType)
+	isGcpSubscriptionType := args.subscriptionType == fmt.Sprintf("%s (%s)", gcpBillingModel.Id(), gcpBillingModel.Description())
```
cristianoveiga, line 515: `isGcpSubscriptionType` not all GCP subscriptions falls into this. I would suggest naming it `isGcpMarketplace` instead.

go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/nwidger/jsoncolor v0.3.2
 	github.com/onsi/ginkgo/v2 v2.11.0
 	github.com/onsi/gomega v1.27.8
-	github.com/openshift-online/ocm-sdk-go v0.1.356
+	github.com/openshift-online/ocm-sdk-go v0.1.367
```
cristianoveiga, line 15: We will need to wait the next version as we are changing the BillingModelItem from a struct to a class. 

See https://github.com/openshift-online/ocm-api-model/pull/831/files

cmd/ocm/create/cluster/cmd.go
```diff
@@ -1036,8 +1137,15 @@ func promptExistingVPC(fs *pflag.FlagSet, connection *sdk.Connection) error {
...
 }
 
-func promptCCS(fs *pflag.FlagSet) error {
-	err := arguments.PromptBool(fs, "ccs")
+func promptCCS(fs *pflag.FlagSet, isGcpSubscriptionType bool) error {
+	var err error
+	// If marketplace-gcp subscription type is used, ccs should by default be true
+	if isGcpSubscriptionType {
+		fmt.Println("setting ccs to 'true'")
+		args.ccs.Enabled = true
```
cristianoveiga, line 1145: We could move this logic outside this function

cristianoveiga (changes requested):
@tirthct  PTAL

cmd/ocm/create/cluster/cmd.go
tirthct: Simplified by using a function to create the option and reusing outside the if else block

cmd/ocm/create/cluster/cmd.go
tirthct: Moved to the billing package

cmd/ocm/create/cluster/cmd.go
tirthct: Create a function and reused wherever duplicated

cmd/ocm/create/cluster/cmd.go
tirthct: Changed

cmd/ocm/create/cluster/cmd.go
tirthct: Moved outside this function and passed a value on whether to prompt for ccs or not

cmd/ocm/create/cluster/cmd.go
tirthct: Used similar methods as subnets

go.mod
tirthct: Used latest sdk version

cmd/ocm/create/cluster/cmd.go
```diff
@@ -19,6 +19,7 @@ package cluster
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/openshift-online/ocm-cli/pkg/billing"
```
cristianoveiga, line 22: order of imports...

tirthct: Fixed

cmd/ocm/create/cluster/cmd.go
cristianoveiga: I think we can simplify further...

How about this:

```
		if billingModel.ID() == billing.StandardSubscriptionType {
			options = append([]arguments.Option{option}, options...)
		} else {
			options = append(options, option)
		}
```

cmd/ocm/create/cluster/cmd.go
```diff
@@ -435,7 +504,12 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		args.clusterWideProxy.Enabled = true
 	}
 
-	err = promptCCS(fs)
+	// If marketplace-gcp subscription type is used, ccs should by default be true
+	if isGcpMarketplaceSubscriptionType {
+		fmt.Println("setting ccs to 'true'")
+		args.ccs.Enabled = true
+	}
+	err = promptCCS(fs, args.ccs.Enabled == true)
```
cristianoveiga, line 512: promptCCS(fs, args.ccs.Enabled == true) can go in a else block here, no ? This way we could avoid the extra parameter.

cmd/ocm/create/cluster/cmd.go
```diff
@@ -301,6 +313,16 @@ func init() {
 		fmt.Sprintf("Namespace Ownership Policy for ingress. Options are %s",
 			strings.Join(ingress.ValidNamespaceOwnershipPolicies, ",")),
 	)
+
+	fs.StringVar(
+		&args.subscriptionType,
+		"subscription-type",
+		billing.StandardSubscriptionType,
+		fmt.Sprintf("The subscription billing model for the cluster. Options are %s",
```
cristianoveiga, line 321: missing punctuation mark

pkg/cluster/cluster.go
```diff
@@ -27,7 +27,7 @@ import (
 	"time"
 
 	sdk "github.com/openshift-online/ocm-sdk-go"
-	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
+	amv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
```
cristianoveiga, line 30: was this intentional ? In the whoami package we are also using amsv1.

cmd/ocm/create/cluster/cmd.go
tirthct: We still need to call promtCCS as "service account file" prompt logic is present over there. And that is based on whether args.ccs.Enables is true or false

pkg/cluster/cluster.go
tirthct: I changed these to keep is consistent with CS (they have used cmv1 instead of cmsv1), but we can revert this back

cmd/ocm/create/cluster/cmd.go
tirthct: Keeping the options description consistent with https://github.com/openshift-online/ocm-cli/blob/main/cmd/ocm/create/cluster/cmd.go#L294 and https://github.com/openshift-online/ocm-cli/blob/main/cmd/ocm/create/cluster/cmd.go#L301

cmd/ocm/create/cluster/cmd.go
cristianoveiga: okay... I was not expecting that.

Should we replace "args.ccs.Enabled == true" with  args.ccs.Enabled ?

tirthct: Simplified

cristianoveiga (approved):
LGTM

tirthct:
@gdbranco can you PTAL?

gdbranco (approved):

PR #552, opened 2023-09-20
OCM-3801 | Feat | Added default values for CIDRs and host prefix

tirthct:
**Changed**
- Fetched default values from flavour endpoint
- Set the values for CIDRs and host prefix

**Tested**
- CIDR and host prefix should have the default values auto populated in case of no other input : tested

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -371,6 +371,34 @@ func getFlavourOptions(connection *sdk.Connection) ([]arguments.Option, error) {
 	return options, nil
 }
 
+func GetDefaultClusterFlavors(connection *sdk.Connection, flavour string) (dMachinecidr *net.IPNet, dPodcidr *net.IPNet,
+	dServicecidr *net.IPNet, dhostPrefix int) {
+	flavourGetResponse, err := connection.ClustersMgmt().V1().Flavours().Flavour(flavour).Get().Send()
+	if err != nil {
+		flavourGetResponse, _ = connection.ClustersMgmt().V1().Flavours().Flavour("osd-4").Get().Send()
+	}
+
+	network, ok := flavourGetResponse.Body().GetNetwork()
+	if !ok {
+		return nil, nil, nil, 0
+	}
+	_, dMachinecidr, err = net.ParseCIDR(network.MachineCIDR())
+	if err != nil {
+		dMachinecidr = nil
+	}
+	_, dPodcidr, err = net.ParseCIDR(network.PodCIDR())
+	if err != nil {
+		dPodcidr = nil
+	}
+	_, dServicecidr, err = net.ParseCIDR(network.ServiceCIDR())
+	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/ocm/create/cluster/cmd.go
```diff
@@ -595,6 +623,11 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		return err
 	}
 
+	if args.interactive {
+		machineCIDR, serviceCIDR, podCIDR, hostPrefix := GetDefaultClusterFlavors(connection, args.flavour)
```
cristianoveiga, line 627: is the order of variables/matching the return values? serviceCIDR and podCIDR seem inverted

pkg/arguments/interactive.go
```diff
@@ -295,12 +295,14 @@ func PromptIPNet(fs *pflag.FlagSet, flagName string) error {
 		if flag.Changed {
 			return nil
 		}
-
+		// If the default value is nil, set the value as a default value
+		if flag.DefValue == "<nil>" {
+			flag.DefValue = flag.Value.String()
+		}
```
cristianoveiga, line 301: It's not clear to me why we need to set the flag default value here, can you explain, please?

pkg/arguments/interactive.go
```diff
@@ -295,12 +295,14 @@ func PromptIPNet(fs *pflag.FlagSet, flagName string) error {
...
 		prompt := &survey.Input{
 			Message: getQuestion(flag),
 			Help:    flag.Usage,
-			// TODO respect flag default, if set
-			// (awkward because https://github.com/golang/go/issues/39516).
```
cristianoveiga, line 303: I would keep the comment with the link to this issue which explains the `nil` issue.

cristianoveiga (changes requested):

pkg/arguments/interactive.go
tirthct: We set the default value so that it is set (and shown) on the console when the user does not input any values. When we set the value in https://github.com/openshift-online/ocm-cli/pull/552/files#diff-2f19822c49dc0bec88a73920ea0f04a30b7047b5a827c90447e02b7378744f97R627, it sets the "value" attribute and not the defValue attribute.

pkg/arguments/interactive.go
tirthct: Added back

cmd/ocm/create/cluster/cmd.go
tirthct: Fixed

pkg/arguments/interactive.go
cristianoveiga: This is the implementation of the // TODO below, right. You can probably move that comment here - it will be more clear.

tirthct: Moved

cristianoveiga (approved):
LGTM

gdbranco (approved):

PR #555, opened 2023-09-25
Fix printing of duplicate machine pool information

ThirumlaDevi:


jhernand (approved):
Looks good to me, thanks @ThirumlaDevi.

tests/list_machinepools_test.go
```diff
@@ -0,0 +1,189 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
```
machi1990, line 2:
```suggestion
Copyright (c) 2023 Red Hat, Inc.
```

miguelsorianod:
The changes looks good to me :+1: .

Some notes: this changes the behavior of the `list machinepools` command compared to previous OCM CLI versions:
* Now when listing a cluster that has only the machinepool automatically created on cluster creation it shows the machine pool id of that cluster, which is named `worker` at the moment of writing this, instead of an entry with the `default` name. This means no `default` entry is showed anymore when listing a cluster that has only the machinepool automatically created on cluster creation
* There is no concept of a "default machinepool" anymore on the list machinepools command. The machinepools command simply outputs the list of machinepools IDs retrieved by Cluster Service. One of the machinepools created in the cluster can have the ID `default` by itself and therefore it would be shown as the result of the `list machinepools` output, although it doesn't mean it is a "default machinepool", it is just the ID.

I understand this behavior change was discussed and accepted by the appropriate parties as discussed. If that were not to be the case please double-check with whoever is required. Thanks.

tests/list_machinepools_test.go
ThirumlaDevi: done

machi1990: lgtm

Thanks

PR #556, opened 2023-09-26
OCM 3768 | Feat | Filtered OCM versions for marketplace gcp clusters

tirthct:
**Changed**
- Added gcpMarketplaceEnabled flag for filtering out supported versions of OCM CLI

**Tested**
- Only supported versions should be visible for gcp marketplace clusters : tested

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -400,16 +401,16 @@ func GetDefaultClusterFlavors(connection *sdk.Connection, flavour string) (dMach
 }
 
 func getVersionOptions(connection *sdk.Connection) ([]arguments.Option, error) {
-	options, _, err := getVersionOptionsWithDefault(connection, "")
+	options, _, err := getVersionOptionsWithDefault(connection, "", nil)
 	return options, err
 }
 
-func getVersionOptionsWithDefault(connection *sdk.Connection, channelGroup string) (
+func getVersionOptionsWithDefault(connection *sdk.Connection, channelGroup string, gcpMarketplaceEnabled *string) (
 	options []arguments.Option, defaultVersion string, err error,
 ) {
 	// Check and set the cluster version
 	versionList, defaultVersion, err := c.GetEnabledVersions(
-		connection.ClustersMgmt().V1(), channelGroup)
+		connection.ClustersMgmt().V1(), channelGroup, gcpMarketplaceEnabled)
 	if err != nil {
 		return
 	}
@@ -551,7 +552,12 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		return err
 	}
 
-	versions, defaultVersion, err := getVersionOptionsWithDefault(connection, args.channelGroup)
+	var gcpMarketplaceEnabled string
+	if isGcpMarketplaceSubscriptionType {
+		gcpMarketplaceEnabled = strconv.FormatBool(isGcpMarketplaceSubscriptionType)
+	}
+	versions, defaultVersion, err := getVersionOptionsWithDefault(connection, args.channelGroup,
+		&gcpMarketplaceEnabled)
 	if err != nil {
 		return err
 	}
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -401,11 +401,11 @@ func GetDefaultClusterFlavors(connection *sdk.Connection, flavour string) (dMach
 }
 
 func getVersionOptions(connection *sdk.Connection) ([]arguments.Option, error) {
-	options, _, err := getVersionOptionsWithDefault(connection, "", nil)
+	options, _, err := getVersionOptionsWithDefault(connection, "", "")
 	return options, err
 }
 
-func getVersionOptionsWithDefault(connection *sdk.Connection, channelGroup string, gcpMarketplaceEnabled *string) (
+func getVersionOptionsWithDefault(connection *sdk.Connection, channelGroup string, gcpMarketplaceEnabled string) (
 	options []arguments.Option, defaultVersion string, err error,
 ) {
 	// Check and set the cluster version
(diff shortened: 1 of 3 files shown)


pkg/cluster/versions.go
```diff
@@ -42,12 +42,12 @@ func EnsureOpenshiftVPrefix(v string) string {
...
+func GetEnabledVersions(client *cmv1.Client, channelGroup string, gcpMarketplaceEnabled string) (
 	versions []string, defaultVersion string, err error) {
 	collection := client.Versions()
 	page := 1
 	size := 100
-	filter := "enabled = 'true'"
+	filter := fmt.Sprintf("enabled = 'true' AND gcp_marketplace_enabled = '%s'", gcpMarketplaceEnabled)
 	if channelGroup != "" {
 		filter = fmt.Sprintf("%s AND channel_group = '%s'", filter, channelGroup)
 	}
```
cristianoveiga, line 53: it would be probably safer to only set the gcp_marketplace_enabled restriction if a value is explicitly  provided (like channel group).

I would keep `filter := "enabled = 'true'"` as the base query only.

cristianoveiga (changes requested):

cmd/ocm/list/version/cmd.go
```diff
@@ -55,6 +56,13 @@ func init() {
 		"stable",
 		"List only versions from the specified channel group",
 	)
+
+	fs.StringVar(
+		&args.gcpMarketplaceEnabled,
+		"gcp-marketplace-enabled",
```
cristianoveiga, line 62: this is a long name for the flag. Maybe `gcp-marketplace` or `marketplace-gcp` (to match the subscription type value) would be enough?

cristianoveiga (changes requested):

cmd/ocm/list/version/cmd.go
tirthct: As per the suggestion (https://redhat-internal.slack.com/archives/C05BJQLPYQ5/p1695828401018519?thread_ts=1695762022.253959&cid=C05BJQLPYQ5), changed the flag to gcp-marketplace

pkg/cluster/versions.go
tirthct: Changed

pkg/cluster/versions.go
```diff
@@ -42,12 +42,13 @@ func EnsureOpenshiftVPrefix(v string) string {
...
 // The returned strings are the IDs without "openshift-v" prefix (e.g. "4.6.0-rc.4-candidate")
 // sorted in approximate SemVer order (handling of text parts is somewhat arbitrary).
-func GetEnabledVersions(client *cmv1.Client, channelGroup string) (
+func GetEnabledVersions(client *cmv1.Client, channelGroup string, gcpMarketplaceEnabled string) (
 	versions []string, defaultVersion string, err error) {
 	collection := client.Versions()
 	page := 1
 	size := 100
 	filter := "enabled = 'true'"
+	filter = fmt.Sprintf("%s AND gcp_marketplace_enabled = '%s'", filter, gcpMarketplaceEnabled)
```
cristianoveiga, line 51: I don't think we should filter for `gcp_marketplace_enabled` as long the flag is explicitly set.

You can probably achieve this by using a pointer and checking for nil.

tirthct: Changed

cmd/ocm/create/cluster/cmd.go
```diff
@@ -400,16 +401,16 @@ func GetDefaultClusterFlavors(connection *sdk.Connection, flavour string) (dMach
 }
 
 func getVersionOptions(connection *sdk.Connection) ([]arguments.Option, error) {
-	options, _, err := getVersionOptionsWithDefault(connection, "")
+	options, _, err := getVersionOptionsWithDefault(connection, "", "false")
```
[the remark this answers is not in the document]
tirthct: Changed to nil

cmd/ocm/list/version/cmd.go
```diff
@@ -55,6 +56,13 @@ func init() {
 		"stable",
 		"List only versions from the specified channel group",
 	)
+
+	fs.StringVar(
+		&args.gcpMarketplace,
+		"gcp-marketplace",
+		"false",
```
cristianoveiga, line 63: is this the default value?

tirthct: Yes, changed to an empty string

cmd/ocm/create/cluster/cmd.go
```diff
@@ -551,7 +552,9 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		return err
 	}
 
-	versions, defaultVersion, err := getVersionOptionsWithDefault(connection, args.channelGroup)
+	gcpMarketplaceEnabled := strconv.FormatBool(isGcpMarketplaceSubscriptionType)
+	versions, defaultVersion, err := getVersionOptionsWithDefault(connection, args.channelGroup,
+		&gcpMarketplaceEnabled)
```
cristianoveiga, line 557: This will cause filtering by gcp_marketplace_enabled = 'false' when selecting billing models other than `marketplace-gcp` - which is not the goal, PTAL

tirthct: Fixed the logic

cmd/ocm/create/cluster/cmd.go
```diff
@@ -400,16 +401,16 @@ func GetDefaultClusterFlavors(connection *sdk.Connection, flavour string) (dMach
 }
 
 func getVersionOptions(connection *sdk.Connection) ([]arguments.Option, error) {
-	options, _, err := getVersionOptionsWithDefault(connection, "")
+	options, _, err := getVersionOptionsWithDefault(connection, "", nil)
```
cristianoveiga, line 404: since the default value of the flag is an empty string, does it make sense to use the same pattern here and change the type from *string to string?

tirthct: Yes, changed to keep a string instead of *string

cmd/ocm/list/version/cmd.go
```diff
@@ -55,6 +56,13 @@ func init() {
 		"stable",
 		"List only versions from the specified channel group",
 	)
+
+	fs.StringVar(
+		&args.gcpMarketplace,
+		"gcp-marketplace",
```
cristianoveiga, line 62: should this be `marketplace-gcp` instead?

cristianoveiga:
LGTM

cmd/ocm/list/version/cmd.go
tirthct: Changed

gdbranco (approved):

PR #566, opened 2023-10-09
OCM 3993 | Feat | Add GCP marketplace terms and conditions for marketplace GCP clusters

tirthct:
**Changed**
- Added terms and conditions flag for GCP
- Added check to confirm the flag is set to true for both interactive and non-interactive mode

**Tested**
- Interactive/Non-interactive mode should create clusters only after the flag is set to true

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -46,6 +46,9 @@ const (
 	defaultIngressExcludedNamespacesFlag       = "default-ingress-excluded-namespaces"
 	defaultIngressWildcardPolicyFlag           = "default-ingress-wildcard-policy"
 	defaultIngressNamespaceOwnershipPolicyFlag = "default-ingress-namespace-ownership-policy"
+	gcpTermsAgreementsHyperlink                = "https://console.cloud.google.com" +
+		"/marketplace/agreements/redhat-marketplace/red-hat-openshift-dedicated"
+	gcpTermsAgreementError = "Please accept Google Terms and Agreements in order to proceed"
 )
 
 var args struct {
@@ -71,6 +74,7 @@ var args struct {
 	gcpServiceAccountFile arguments.FilePath
 	etcdEncryption        bool
 	subscriptionType      string
+	marketplaceGcpTerms   bool
 
 	// Scaling options
 	computeMachineType string
@@ -324,6 +328,15 @@ func init() {
 	)
 	arguments.SetQuestion(fs, "subscription-type", "Subscription type:")
 	Cmd.RegisterFlagCompletionFunc("subscription-type", arguments.MakeCompleteFunc(getSubscriptionTypeOptions))
+
+	fs.BoolVar(
+		&args.marketplaceGcpTerms,
+		"marketplace-gcp-terms",
+		false,
+		fmt.Sprintf("Review and accept Google Terms and Agreements on %s. "+
+			"Set the flag to true once agreed in order to proceed further.", gcpTermsAgreementsHyperlink),
+	)
+	arguments.SetQuestion(fs, "marketplace-gcp-terms", "I have accepted Google Terms and Agreements:")
 }
 
 func osdProviderOptions(_ *sdk.Connection) ([]arguments.Option, error) {
@@ -516,6 +529,18 @@ func preRun(cmd *cobra.Command, argv []string) error {
 	if isGcpMarketplace {
 		fmt.Println("setting provider to", c.ProviderGCP)
 		args.provider = c.ProviderGCP
+		fmt.Println("setting ccs to 'true'")
+		args.ccs.Enabled = true
+		if args.interactive {
+			fmt.Println("Review and accept Google Terms and Agreements on", gcpTermsAgreementsHyperlink)
+			err = arguments.PromptBool(fs, "marketplace-gcp-terms")
+			if err != nil {
+				return err
+			}
+		}
+		if !args.marketplaceGcpTerms {
+			return fmt.Errorf(gcpTermsAgreementError)
+		}
 	} else {
 		err = arguments.PromptOneOf(fs, "provider", providers)
 		if err != nil {
@@ -529,11 +554,6 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		args.clusterWideProxy.Enabled = true
 	}
 
-	// If marketplace-gcp subscription type is used, ccs should by default be true
-	if isGcpMarketplace {
-		fmt.Println("setting ccs to 'true'")
-		args.ccs.Enabled = true
-	}
 	err = promptCCS(fs, args.ccs.Enabled)
 	if err != nil {
 		return err


cmd/ocm/create/cluster/cmd.go
```diff
@@ -533,6 +546,14 @@ func preRun(cmd *cobra.Command, argv []string) error {
 	if isGcpMarketplace {
```
cristianoveiga, line 546: We are doing the same check on line 529, I wonder if this can be refactored so that we only check once.

cmd/ocm/create/cluster/cmd.go
```diff
@@ -533,6 +546,14 @@ func preRun(cmd *cobra.Command, argv []string) error {
...
 		fmt.Println("setting ccs to 'true'")
 		args.ccs.Enabled = true
+		fmt.Println("Review and accept Google Terms and Agreements on", gcpTermsAgreementsHyperlink)
```
cristianoveiga, line 549: should it be `Terms and Conditions` instead?

cmd/ocm/create/cluster/cmd.go
```diff
@@ -46,6 +46,9 @@ const (
 	defaultIngressExcludedNamespacesFlag       = "default-ingress-excluded-namespaces"
 	defaultIngressWildcardPolicyFlag           = "default-ingress-wildcard-policy"
 	defaultIngressNamespaceOwnershipPolicyFlag = "default-ingress-namespace-ownership-policy"
+	gcpTermsAgreementsHyperlink                = "https://console.cloud.google.com" +
+		"/marketplace/agreements/redhat-marketplace/red-hat-openshift-dedicated"
+	gcpTermsAgreementError = "Please agree to Google Terms and Agreements in order to proceed"
```
cristianoveiga, line 51: Same question here: should it be terms and conditions instead?

cmd/ocm/create/cluster/cmd.go
```diff
@@ -324,6 +328,15 @@ func init() {
 	)
 	arguments.SetQuestion(fs, "subscription-type", "Subscription type:")
 	Cmd.RegisterFlagCompletionFunc("subscription-type", arguments.MakeCompleteFunc(getSubscriptionTypeOptions))
+
+	fs.BoolVar(
+		&args.marketplaceGcpTerms,
+		"marketplace-gcp-terms",
+		false,
```
cristianoveiga, line 335: should the default value be true, instead?

cmd/ocm/create/cluster/cmd.go
```diff
@@ -668,6 +689,17 @@ func run(cmd *cobra.Command, argv []string) error {
 		args.subscriptionType = parseSubscriptionType(args.subscriptionType)
 	}
 
+	// If marketplace-gcp subscription type is used, flag marketplace-gcp-terms must be set to true.
+	gcpBillingModel, _ := billing.GetBillingModel(connection, billing.MarketplaceGcpSubscriptionType)
+	gcpSubscriptionTypeTemplate := parseSubscriptionType(subscriptionTypeOption(gcpBillingModel.ID(),
+		gcpBillingModel.Description()).Value)
+	isGcpMarketplaceSubscriptionType := args.subscriptionType == gcpSubscriptionTypeTemplate
+	if isGcpMarketplaceSubscriptionType {
```
cristianoveiga, line 697: nit: isGcpMarketplaceSubscriptionType is a long name for a variable. I would suggest using `isGcpMarketplace`.

cmd/ocm/create/cluster/cmd.go
tirthct: By default we are setting it to false so that the user can override it to true when the terms and conditions are accepted. Otherwise they can skip this without noticing if it is set to true

cmd/ocm/create/cluster/cmd.go
tirthct: Keeping agreements for consistency : https://docs.google.com/document/d/1xpRCZ0QYWRuvVtCEycWrHwXWSIepzUkgb8Nobqf3Dlc/edit#heading=h.ve1e69gtx5pl

cmd/ocm/create/cluster/cmd.go
```diff
@@ -668,6 +689,17 @@ func run(cmd *cobra.Command, argv []string) error {
...
+		if !args.marketplaceGcpTerms {
+			return fmt.Errorf(gcpTermsAgreementError)
+		}
+	}
+
```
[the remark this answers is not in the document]
tirthct: Combined the check in preRun

cmd/ocm/create/cluster/cmd.go
cristianoveiga: we are checking if `isGcpMarketplace` twice, lines 529 and 548. I'm wondering if we could "merge" the two code blocks so that the logic is centralized.

cmd/ocm/create/cluster/cmd.go
cristianoveiga: yes, that makes sense to me. Thanks for clarifying.

cmd/ocm/create/cluster/cmd.go
tirthct: Merged

cmd/ocm/create/cluster/cmd.go
cristianoveiga: The console says "accept" instead of agree, should we update here too?

cmd/ocm/create/cluster/cmd.go
```diff
@@ -514,7 +527,20 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		parseSubscriptionType(args.subscriptionType) == parseSubscriptionType(gcpSubscriptionTypeTemplate.Value)
 
 	if isGcpMarketplace {
-		fmt.Println("setting provider to", c.ProviderGCP)
+		if args.interactive {
+			fmt.Println("setting provider to", c.ProviderGCP)
+			fmt.Println("setting ccs to 'true'")
```
cristianoveiga, line 532: this change the behaviour: why do we only show the message/override the ccs and cloud provider attributes in the interactive mode?

cmd/ocm/create/cluster/cmd.go
```diff
@@ -514,7 +527,20 @@ func preRun(cmd *cobra.Command, argv []string) error {
...
+			err = arguments.PromptBool(fs, "marketplace-gcp-terms")
+			if err != nil {
+				return err
+			}
+		}
+		if !args.marketplaceGcpTerms {
+			return fmt.Errorf(gcpTermsAgreementError)
+		}
+		args.ccs.Enabled = true
 		args.provider = c.ProviderGCP
```
cristianoveiga, line 544: I would have these assignments close  to the code where we print messages related to the override

cmd/ocm/create/cluster/cmd.go
tirthct: Updated

cmd/ocm/create/cluster/cmd.go
tirthct: Added the prompt messages for non-interactive mode as well

cmd/ocm/create/cluster/cmd.go
tirthct: Added them after the prompt messages

cristianoveiga:
LGTM

cristianoveiga (approved):

PR #567, opened 2023-10-10
OCM-4183 | feat: suggest pop out an error message instead of stop the progress if user do not input the service account file

cristianoveiga:
…s not provided


tirthct (approved):

tirthct:
LGTM

PR #568, opened 2023-10-10
OCM-4184 | Feat | Convert relative path containing tilde for service account file

tirthct:
**Changed**
- Convert relative path containing tilde for service account file

**Tested**
- Path containing tilde should be valid for service account file

### DIFF
--- a/pkg/arguments/interactive.go
+++ b/pkg/arguments/interactive.go
@@ -301,12 +301,10 @@ func resolveRelativePath(path string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	dir := usr.HomeDir
-	absPath := dir
-	if strings.HasPrefix(path, "~/") && len(path) > 2 {
-		absPath = filepath.Join(dir, path[2:])
+	if strings.HasPrefix(path, "~/") {
+		path = strings.Replace(path, "~", usr.HomeDir, 1)
 	}
-	return absPath, nil
+	return path, nil
 }
 
 // PromptIPNet sets an optional IPNet flag value interactively, unless already set.
--- a/pkg/arguments/interactive.go
+++ b/pkg/arguments/interactive.go
@@ -294,6 +294,9 @@ func PromptFilePath(fs *pflag.FlagSet, flagName string, required bool) error {
 // Golang does not support tilde file paths https://github.com/golang/go/issues/57569
 // However, we try to resolve this by manually so user can proceed further
 func resolveRelativePath(path string) (string, error) {
+	if !strings.Contains(path, "~") {
+		return path, nil
+	}
 	usr, err := user.Current()
 	if err != nil {
 		return "", err
--- a/pkg/arguments/interactive.go
+++ b/pkg/arguments/interactive.go
@@ -299,15 +299,11 @@ func resolveRelativePath(path string) (string, error) {
 		return "", err
 	}
 	dir := usr.HomeDir
-	if path == "~" {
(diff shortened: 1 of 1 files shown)


pkg/arguments/interactive.go
```diff
@@ -280,12 +281,35 @@ func PromptFilePath(fs *pflag.FlagSet, flagName string, required bool) error {
...
+	}
+	dir := usr.HomeDir
+	if path == "~" {
+		// In case of "~", "else if" would not catch that
+		path = dir
+	} else if strings.HasPrefix(path, "~/") {
+		// Use strings.HasPrefix so we don't match paths like
+		// "/something/~/something/"
+		path = filepath.Join(dir, path[2:])
+	}
```
cristianoveiga, line 309: can we simplify this logic?

tirthct: Removed the earlier if condition and added assignment to a new variable

pkg/arguments/interactive.go
```diff
@@ -280,12 +281,31 @@ func PromptFilePath(fs *pflag.FlagSet, flagName string, required bool) error {
...
+func resolveRelativePath(path string) (string, error) {
+	usr, err := user.Current()
+	if err != nil {
+		return "", err
+	}
+	dir := usr.HomeDir
+	absPath := dir
+	if strings.HasPrefix(path, "~/") && len(path) > 2 {
+		absPath = filepath.Join(dir, path[2:])
+	}
```
cristianoveiga, line 305: how about a direct string replacement?

As in:

`strings.Replace(path, "~", usr.HomeDir, 1)`

pkg/arguments/interactive.go
```diff
@@ -280,12 +281,31 @@ func PromptFilePath(fs *pflag.FlagSet, flagName string, required bool) error {
...
 			fs.Set(flagName, response)
 		}
 		return nil
 	})
 }
 
+// Golang does not support tilde file paths https://github.com/golang/go/issues/57569
+// However, we try to resolve this by manually so user can proceed further
+func resolveRelativePath(path string) (string, error) {
+	usr, err := user.Current()
```
cristianoveiga, line 297: The majority of calls won't have the ~.

We could add a check and return earlier

tirthct: Added

pkg/arguments/interactive.go
cristianoveiga: IMO, that would be an user input error and there's no much we could do anyway - unless I am missing something. 

Can you explain/provide a valid user case like that?

tirthct: Linux allows creating folders with a tilde sign in between. The user may have the service account file somewhere in the folder, and we should not restrict them from using that. With this change, we will be confusing the user with file not specified error.

cristianoveiga: I see. Thanks for the example. 

I think we would be still covered with my suggested code, since only one replacement would be done:

```
	if strings.HasPrefix(path, "~/") {
		absPath = strings.Replace(path, "~", usr.HomeDir, 1)
	}
```

wdyt?

tirthct: Oh you mean keeping the if condition and then replacing! The UI showed me all the lines in the comment for some reason so I thought you suggested replacing all of it. 
But this change works. Updated the code.

cristianoveiga (approved):

PR #574, opened 2023-10-31
Add vpc-project-id flag for shared VPC on GCP

JakobGray:
Issue: https://issues.redhat.com/browse/OCM-4465

Adds `--vpc-project-id` flag to specify host project id when creating a shared-vpc GCP cluster. 

In the interactive flow, if the user selects yes to `Install into an existing VPC` they will get a follow-up prompt `Install into a shared VPC`, which defaults to No.

```
? Install into an existing VPC (optional): Yes
? vpc name: <vpc-name>
? control plane subnet: <subnet-name>
? compute subnet: <subnet-name>
? Install into a shared VPC (optional): Yes
? Vpc project id: <project-name>
```

In the existing flow when `use-existing-vpc` is true the cli will run client-side validation that the vpc and subnets exist. If using shared vpc this validation will be skipped (because it can't be verified with existing credentials).

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -1062,7 +1062,7 @@ func cleanSecurityGroups(securityGroups *[]string) {
 
 func wasGCPNetworkReceived() bool {
 	return args.existingVPC.VPCName != "" && args.existingVPC.ControlPlaneSubnet != "" &&
-		args.existingVPC.ComputeSubnet != ""
+		args.existingVPC.ComputeSubnet != "" && args.existingVPC.VPCProjectID != ""
 }
 
 func promptExistingGCPVPC(fs *pflag.FlagSet, connection *sdk.Connection) error {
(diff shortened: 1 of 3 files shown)


cmd/ocm/create/cluster/cmd.go
```diff
@@ -1054,7 +1054,7 @@ func promptExistingAWSVPC(fs *pflag.FlagSet, connection *sdk.Connection) error {
 
 func wasGCPNetworkReceived() bool {
 	return args.existingVPC.VPCName != "" && args.existingVPC.ControlPlaneSubnet != "" &&
-		args.existingVPC.ComputeSubnet != ""
+		args.existingVPC.ComputeSubnet != "" && args.existingVPC.VPCProjectID != ""
```
ckandag, line 1057: @JakobGray   vpcProjectID is optional.   Incase of a Non-XPN existing VPC we wouldnt set VPCProjectID.  I dont think we should be doing a "&&" here

pkg/cluster/cluster.go
```diff
@@ -487,7 +491,12 @@ func CreateCluster(cmv1Client *cmv1.Client, config Spec, dryRun bool) (*cmv1.Clu
 }
 
 func isGCPNetworkExists(existingVPC ExistingVPC) bool {
-	return existingVPC.VPCName != "" || existingVPC.ControlPlaneSubnet != "" || existingVPC.ComputeSubnet != ""
+	return existingVPC.VPCName != "" || existingVPC.ControlPlaneSubnet != "" ||
```
ckandag, line 494: Optional:  Is checking for existence one of every vpc field correct here ?

Looks like the expectation is that input validation is done by the time this is called.
I think it should suffice if one of the fields i.e  VPCName  exists  or Check  every vpc related field is non-empty before setting it on the clusterBuilder

cmd/ocm/create/cluster/cmd.go
```diff
@@ -1071,22 +1071,47 @@ func promptExistingGCPVPC(fs *pflag.FlagSet, connection *sdk.Connection) error {
 		}
 	}
 
-	if args.existingVPC.Enabled || wasGCPNetworkReceived() {
-		err = arguments.PromptString(fs, "vpc-name")
-		if err != nil {
-			return err
-		}
+	if !args.existingVPC.Enabled && !wasGCPNetworkReceived() {
```
ckandag, line 1074: What happens if interactive is enabled but the top level EnableExistingVPC is not enabled.

Will it be caught somewhere else and prompt user for enabling existing VPC

pkg/cluster/cluster.go
ckandag: After looking through the code some more , I think the check should simply be   exisitingVPC.enabled

cmd/ocm/create/cluster/cmd.go
ckandag: nm, I see the interactive prompt for the EnableExistingVPC right   above

cmd/ocm/create/cluster/cmd.go
```diff
@@ -1140,27 +1165,31 @@ func promptExistingGCPVPC(fs *pflag.FlagSet, connection *sdk.Connection) error {
...
+	}
+	flag = fs.Lookup("control-plane-subnet")
+	if !flag.Changed {
+		fs.Set("control-plabe-subnet", args.existingVPC.ControlPlaneSubnet)
+	}
+	flag = fs.Lookup("compute-subnet")
+	if !flag.Changed {
+		fs.Set("compute-subnet", args.existingVPC.ComputeSubnet)
+	}
+	flag = fs.Lookup("vpc-project-id")
```
ckandag, line 1187: Is all this fs.Set really necessary ? PromptString appears to do that already.

pkg/cluster/cluster.go
JakobGray: `exisitingVPC.enabled` is a derived value that may not be set. I see no reason to change the existing function implementation.

renan-campos (approved):
/lgtm

PR #576, opened 2023-11-12
to update to golang1.21.3

sdminonne:
Since 1.18 started to be pretty updated


andreadecorte:
outdated - we are at 1.23 now.

PR #581, opened 2023-12-01
OCM-5131 | feat: add SG IDs to describe cluster and list machinepools

gdbranco:
Related issue: OCM-5131


cmd/ocm/list/machinepool/cmd.go
```diff
@@ -132,6 +133,13 @@ func printReplicas(autoscaling *cmv1.MachinePoolAutoscaling, replicas int) strin
 	return fmt.Sprintf("%d", replicas)
 }
 
+func printAdditionalSecurityGroups(securityGroups []string) string {
+	if len(securityGroups) == 0 {
+		return ""
+	}
```
mnecas, line 139: When the list is empty the join returns an empty string

pkg/cluster/describe.go
```diff
@@ -294,6 +295,15 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
 	return nil
 }
 
+func printNodeInfo(replicasInfo string, securityGroups []string) string {
+	nodeStr := ""
+	nodeStr += fmt.Sprintf("\tReplicas: %s", replicasInfo)
+	if len(securityGroups) > 0 {
+		nodeStr += fmt.Sprintf("\n\tAWS Additional Security Group IDs: %s", strings.Join(securityGroups, ", "))
+	}
```
mnecas, line 303: if there will be security groups the multiple calls of `printNodeInfo` one right after the other will build the string
`"\tReplicas: %s"\tReplicas: %s"\tReplicas: %s` without creating new line. I would suggest to move the `\n` to the end of each string making: `\tReplicas: %s\n` and `\tAWS Additional Security Group IDs: %s\n`

pkg/cluster/describe.go
```diff
@@ -217,9 +218,9 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
 		apiURL,
 		apiListening,
 		cluster.Console().URL(),
-		cluster.Nodes().Master(),
-		cluster.Nodes().Infra(),
-		computesStr,
+		printNodeInfo(fmt.Sprintf("%d", cluster.Nodes().Master()), cluster.AWS().AdditionalControlPlaneSecurityGroupIds()),
+		printNodeInfo(fmt.Sprintf("%d", cluster.Nodes().Infra()), cluster.AWS().AdditionalInfraSecurityGroupIds()),
```
mnecas, line 222: Why not the use of `strconv.Itoa()` instead of the string formatting

pkg/cluster/describe.go
```diff
@@ -294,6 +295,15 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
 	return nil
 }
 
+func printNodeInfo(replicasInfo string, securityGroups []string) string {
+	nodeStr := ""
+	nodeStr += fmt.Sprintf("\tReplicas: %s", replicasInfo)
```
mnecas, line 300: `nodeStr := fmt.Sprintf("\tReplicas: %s", replicasInfo)`

cmd/ocm/list/machinepool/cmd.go
gdbranco: Shouldn't that be the expectation?

pkg/cluster/describe.go
gdbranco: Mostly because this was the pattern prior, adjusted for compute replicas too

pkg/cluster/describe.go
gdbranco: Good catch, I was building things differently at some point

cmd/ocm/list/machinepool/cmd.go
mnecas: ah sorry I did not explain myself properly, my point was that you don't need this part you can use just the join, you don't need this check

gdbranco: Oh I see, interesting. There are lots of instances where we do this :laughing:

pkg/cluster/describe.go
```diff
@@ -294,6 +296,14 @@ func PrintClusterDescription(connection *sdk.Connection, cluster *cmv1.Cluster)
 	return nil
 }
 
+func printNodeInfo(replicasInfo string, securityGroups []string) string {
```
mnecas, line 299: just nit pick what I just noticed, the function does not print but formats so might be better name `formatNodeInfo`

gdbranco: I agree, but it's the current standard for all other functions that are formatting is also to print :laughing:

ciaranRoche (approved):
/lgtm

PR #586, opened 2023-12-12
OCM-5281 | Feat | Add region validation from ocm-shards and list regions command

tirthct:
**Changed**
- Modified ocm login command to pass the region name (with or without rh prefix)
- Add package to get and validate regions
- Add list regions command

**Tested**
- Region name passed in should be validated from ocm-shards
- list rh-regions should show all available region names


cmd/ocm/login/cmd.go
```diff
@@ -265,9 +275,14 @@ func run(cmd *cobra.Command, argv []string) error {
 		cfg.User = ""
 		cfg.Password = ""
 	}
-	err = config.Save(cfg)
-	if err != nil {
-		return fmt.Errorf("Can't save config file: %v", err)
+
+	// If an OCM region is provided, update the config URL with the SDK generated URL
+	if cfg.RhRegion != "" {
+		regValue, err := rhRegion.GetRhRegion(context.Background(), refreshToken, args.rhRegion)
```
macgregor, line 281: so i think region discovery inside login creates a awkward scenario if you try to use `--rh-region` without an active login session. So like say i have never logged in ever. I run `ocm login --rh-region=singapore` and im met with an error "Not logged in, run the 'login' command" well thats what im trying to do...

ocm-shards.json is public, we dont actually need credentials to fetch it, but we do need a url to fetch it from (e.g. api.openshift.com/.../ocm-shards.json for prod or api.stage.openshift.com/.../ocm-shards.json for stage, etc). 

I wonder how we could make it friendlier...

cmd/ocm/list/cmd.go
```diff
@@ -46,4 +47,5 @@ func init() {
 	Cmd.AddCommand(upgradepolicy.Cmd)
 	Cmd.AddCommand(user.Cmd)
 	Cmd.AddCommand(version.Cmd)
+	Cmd.AddCommand(rhRegion.Cmd)
```
macgregor, line 50: is there a way to make the command hidden like the `--rh-region` flag?

pkg/config/config.go
```diff
@@ -48,6 +48,7 @@ type Config struct {
 	RefreshToken string   `json:"refresh_token,omitempty" doc:"Offline or refresh token."`
 	Scopes       []string `json:"scopes,omitempty" doc:"OpenID scope. If this option is used it will replace completely the default scopes. Can be repeated multiple times to specify multiple scopes."`
 	TokenURL     string   `json:"token_url,omitempty" doc:"OpenID token URL."`
+	RhRegion     string   `json:"rh_region,omitempty" doc:"OCM region name."`
```
macgregor, line 51: i dont think we need to store this in the config since its implicit in the url. `ocm login --rh-region=singapore` results in the "singapore" region ocm instance url being stored in URL config variable. We want to try to limit region awareness to just the login command for now until we find a good need to explicitly store the selected region for use in other parts of the cli.

pkg/rhRegion/rhRegion.go
```diff
@@ -0,0 +1,50 @@
+package rhRegion
```
macgregor, line 1: so this whole file is a great example of what can get pushed down into the ocm-sdk for reuse in rosa cli. not necessarily now, but keep it in mind.

pkg/rhRegion/rhRegion.go
```diff
@@ -0,0 +1,50 @@
...
+
+type Region struct {
+	URL string
+	AWS []string
+	GCP []string
+}
+
+func GetRhRegions(ctx context.Context, token string) (map[string]Region, error) {
+	var regions map[string]Region
+	globalConn, err := sdk.NewConnectionBuilder().Tokens(token).RhRegion("integration").BuildContext(ctx)
```
macgregor, line 20: couple things:
1. ocm-shards.json is public, its doesnt actually need a token, you just need a url
2. forget about RhRegion on the connection builder, it should probably be removed, it doesnt make sense there

So see if you can make this function signature work `func GetRhRegions(ctx context.Context, URL String) (map[string]Region, error)`. And then later we can try to refactor `GetRhRegions` and `GetRhRegion` down into the ocm-sdk itself.

pkg/rhRegion/rhRegion.go
```diff
@@ -0,0 +1,50 @@
...
+
+	return regions, nil
+}
+
+func GetRhRegion(ctx context.Context, token string, regionName string) (Region, error) {
+	regions, err := GetRhRegions(ctx, token)
+	if err != nil {
+		return Region{}, err
+	}
+	if !strings.HasPrefix(regionName, "rh-") {
```
macgregor, line 42: I understand the impulse to be flexible or helpful, but in general try to avoid "smart keys" where you support multiple key variations with application logic, it ends up being a pain in the butt to maintain long term and is generally unintuitive to users. 

Speaking from experience :)

cmd/ocm/login/cmd.go
tirthct: If we pass ocm login --rh-region=singapore, I get this message currently:
```
In order to log in it is mandatory to use '--token', '--user' and '--password', or '--client-id' and '--client-secret'.
You can obtain a token at: https://console.redhat.com/openshift/token .
See 'ocm login --help' for full help.
```

Also, token is mandatory right now if I want to create an SDK connection using ConnectionBuilder i.e. 

> globalConn, err := sdk.NewConnectionBuilder().Tokens(token).RhRegion("integration").BuildContext(ctx)

If I try using this command without the token, I get an error 
`either a token, an user name and password or a client identifier and secret are necessary, but none has been provided`

We can use the direct URLs (i.e. instead of sdk just call the api.openshift.com/.../ocm-shards.json directly, but since this might move to SDK, not sure if this wold be a good idea or not

cmd/ocm/list/cmd.go
tirthct: Hid the command

pkg/config/config.go
tirthct: Removed the JSON representation. Should we omit this?

pkg/rhRegion/rhRegion.go
tirthct: Ack.

pkg/rhRegion/rhRegion.go
tirthct: Removed the rh- prefix check

pkg/rhRegion/rhRegion.go
tirthct: Changed sdk call to simple http get method

pkg/rhRegion/rhRegion.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return "", err
+	}
+	regionDiscoveryHost := "api.openshift.com"
+	if strings.HasSuffix(baseUrl.Hostname(), "integration.openshift.com") {
+		regionDiscoveryHost = "api.integration.openshift.com"
+	} else if strings.HasSuffix(baseUrl.Hostname(), "integration.openshift.com") {
+		regionDiscoveryHost = "api.stage.openshift.com"
+	}
+	regionDiscoveryHost = "api.integration.openshift.com"
```
macgregor, line 58: this temporary while we play in integration? add a `//TODO: remove before merge` if so

tirthct: Added TODO

cmd/ocm/login/cmd.go
```diff
@@ -239,6 +247,7 @@ func run(cmd *cobra.Command, argv []string) error {
 
 	// Update the configuration with the values given in the command line:
 	cfg.TokenURL = tokenURL
+	cfg.RhRegion = args.rhRegion
```
macgregor, line 250: again, no need to persist the `RhRegion` to the config. We go from args.rhRegion -> region discover -> cfg.URL, once we have the URL we are done with region. If we find a specific need later where we need to keep track of the name of the region for some purpose, we can add it to the persisted config then. For now, I dont see it being needed after line 284 (`cfg.URL = regValue.URL`)

pkg/config/config.go
```diff
@@ -48,9 +48,10 @@ type Config struct {
 	RefreshToken string   `json:"refresh_token,omitempty" doc:"Offline or refresh token."`
 	Scopes       []string `json:"scopes,omitempty" doc:"OpenID scope. If this option is used it will replace completely the default scopes. Can be repeated multiple times to specify multiple scopes."`
 	TokenURL     string   `json:"token_url,omitempty" doc:"OpenID token URL."`
-	URL          string   `json:"url,omitempty" doc:"URL of the API gateway. The value can be the complete URL or an alias. The valid aliases are 'production', 'staging' and 'integration'."`
-	User         string   `json:"user,omitempty" doc:"User name."`
-	Pager        string   `json:"pager,omitempty" doc:"Pager command, for example 'less'. If empty no pager will be used."`
+	RhRegion     string
```
macgregor, line 51: kill it

pkg/config/config.go
```diff
@@ -270,6 +271,10 @@ func (c *Config) Connection() (connection *sdk.Connection, err error) {
 	}
 	builder.Insecure(c.Insecure)
 
+	if c.RhRegion != "" {
+		builder.RhRegion(c.RhRegion)
```
macgregor, line 275: kill it. We connect to the correct region because cfg.URL is pointed to the correct region url. The sdk ConnectionBuilder will not be aware of the concept of a region.

pkg/rhRegion/rhRegion.go
```diff
@@ -0,0 +1,61 @@
...
+func DetermineRegionDiscoveryUrl(ocmServiceUrl string) (string, error) {
+	baseUrl, err := url.Parse(ocmServiceUrl)
+	if err != nil {
+		return "", err
+	}
+	regionDiscoveryHost := "api.openshift.com"
+	//TODO: Remove the OR condition from this if statement before the MR merge
+	if strings.HasSuffix(baseUrl.Hostname(), "integration.openshift.com") || true {
+		regionDiscoveryHost = "api.integration.openshift.com"
+	} else if strings.HasSuffix(baseUrl.Hostname(), "integration.openshift.com") {
```
macgregor, line 57: stage.openshift.com

cmd/ocm/login/cmd.go
```diff
@@ -265,6 +274,16 @@ func run(cmd *cobra.Command, argv []string) error {
 		cfg.User = ""
 		cfg.Password = ""
 	}
+
+	// If an OCM region is provided, update the config URL with the SDK generated URL
+	if cfg.RhRegion != "" {
+		regValue, err := rhRegion.GetRhRegion(connection.URL(), args.rhRegion)
+		if err != nil {
+			return fmt.Errorf("Can't find region: %v", err)
```
macgregor, line 282: nit: fyi, you should wrap errors with `%w` rather than `%v`. `%w` provides more functionality and preserves stack traces better. people do it incorrectly all over our code base so i point it out when i see it: https://go.dev/blog/go1.13-errors

cmd/ocm/list/rhRegion/cmd.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	connection, err := ocm.NewConnection().Build()
+	if err != nil {
+		return err
+	}
+	defer connection.Close()
+
+	// Check that the configuration has credentials or tokens that don't have expired:
+	armed, reason, err := cfg.Armed()
```
macgregor, line 37: i dont think we need the armed check or an ocm-sdk ConnectionBuilder since ocm-shards is public (anonymous). so i think we can boil this command down to just:
```

	// Load the configuration file:
	cfg, err := config.Load()
	if err != nil {
		return fmt.Errorf("Can't load config file: %v", err)
	}
	if cfg == nil {
		return fmt.Errorf("Not logged in, run the 'login' command")
	}

	regions, err := rhRegion.GetRhRegions(cfg.URL)
	if err != nil {
		return fmt.Errorf("Failed to get OCM regions: %v", err)
	}

	for regionName := range regions {
		fmt.Println(regionName)
	}
	return nil
```

cmd/ocm/login/cmd.go
tirthct: Ack. Removed

go.mod
```diff
@@ -72,9 +73,10 @@ require (
...
 	github.com/zgalor/weberr v0.7.0 // indirect
 	golang.org/x/crypto v0.1.0 // indirect
-	golang.org/x/net v0.10.0 // indirect
 	golang.org/x/sys v0.10.0 // indirect
 	golang.org/x/tools v0.9.3 // indirect
 	google.golang.org/protobuf v1.28.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace github.com/openshift-online/ocm-sdk-go v0.1.388 => ../ocm-sdk-go
```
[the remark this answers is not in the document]
tirthct: Removed since we no longer need RhRegion

pkg/config/config.go
tirthct: Removed

pkg/config/config.go
tirthct: Removed

pkg/rhRegion/rhRegion.go
tirthct: Fixed

cmd/ocm/login/cmd.go
tirthct: Ahh, I see! Changed to %w

cmd/ocm/list/rhRegion/cmd.go
tirthct: Ack. Removed the check

macgregor:
lgtm, thanks Tirth!

macgregor:
lgtm

cmd/ocm/list/cmd.go
```diff
@@ -46,4 +47,7 @@ func init() {
 	Cmd.AddCommand(upgradepolicy.Cmd)
 	Cmd.AddCommand(user.Cmd)
 	Cmd.AddCommand(version.Cmd)
+	rhRegionCmd := rhRegion.Cmd
+	rhRegionCmd.Hidden = true
```
markturansky, line 51: should this be set by default in the cmd?  seems odd to toggle off here.

tirthct: We don't want the command `ocm list rh-regions` to be visible. We can hide the flags but this is the only way to hide the subcommands : https://github.com/spf13/cobra/issues/1507#issuecomment-1632443187

markturansky: if this is always hidden by default, i'm suggesting you default to hidden/true in the cmd itself.  it seems weird to toggle at this level.

tirthct: Ack. Added hidden in Cmd itself

markturansky (approved):
LGTM

markturansky (approved):
LGTM

cmd/ocm/list/rhRegion/cmd.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/openshift-online/ocm-cli/pkg/config"
+	"github.com/spf13/cobra"
+)
+
+var Cmd = &cobra.Command{
+	Use:     "rh-regions",
+	Short:   "List available OCM regions",
+	Long:    "List available OCM regions",
+	Example: `  # List all supported OCM regions ocm list rh-regions`,
```
cristianoveiga, line 15: should this be `ocm list rh-regions` only?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift-online/ocm-cli
 
-go 1.18
+go 1.21
```
cristianoveiga, line 3: is the go version update required here? I'm wondering if it can/should be handled in a separate MR.

tirthct: OCM SDK uses latest golang : https://github.com/openshift-online/ocm-sdk-go/commit/1252182841bbe28fc65153d4c502d95a67b158cf#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6
We can create a separate MR here though.

cmd/ocm/list/rhRegion/cmd.go
tirthct: Example sections in other files have the same format : https://github.com/openshift-online/ocm-cli/blob/main/cmd/ocm/list/version/cmd.go#L38

cristianoveiga: Okay, but IMO that's not really helpful.

cristianoveiga (approved):
LGTM

PR #588, opened 2023-12-20
OCM-4783 | feat: display warnings after cluster creation

renan-campos:
Demo:
```
[rcampos@rcampos-thinkpadt14sgen2i ocm-cli]$ ./ocm create cluster -i
? cluster name rc-test
? Subscription type: standard (Annual: Fixed capacity subscription from Red Hat)
? Cloud provider: gcp
? CCS: Yes
? Service account file: REDACTED
? Multiple AZ: No
? Secure boot support for Shielded VMs: No
? Region: us-east4
? OpenShift version: 4.14.6
? Compute machine type: custom-4-16384
? Enable autoscaling: No
? Compute nodes: 2
? Install into an existing VPC (optional): No
? Machine CIDR: 10.0.0.0/16
? Service CIDR: 172.30.0.0/16
? Pod CIDR: 10.128.0.0/14
? Host prefix: 23
? Private cluster (optional): No

ID:                     REDACTED
External ID:
Name:                   rc-test
Display Name:           rc-test
State:                  validating
API URL:
API Listening:          external
Console URL:
Control Plane:
        Replicas: 3
Infra:
        Replicas: 2
Compute:
        Replicas: 2
Product:                osd
Subscription type:      standard
Provider:               gcp
Version:
Region:                 us-east4
Multi-az:               false
CCS:                    true
HCP:                    false
Subnet IDs:             []
PrivateLink:            false
STS:                    false
Existing VPC:           unsupported
Channel Group:          stable
Cluster Admin:          true
Organization:           REDACTED
Creator:                REDACTED
Email:                  rcampos@redhat.com
AccountNumber:          REDACTED
Created:                2023-12-20T19:58:20Z
Shard:                  REDACTED

⚠️ WARNING:
Please enable the Org Policy API for the GCP project 'REDACTED'
Without the GCP Org Policy API enabled, OCM is unable to determine whether the GCP project contains any policies that would affect the in
stallation
```

### DIFF
--- a/pkg/cluster/describe.go
+++ b/pkg/cluster/describe.go
@@ -25,6 +25,7 @@ import (
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	amv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	slv1 "github.com/openshift-online/ocm-sdk-go/servicelogs/v1"
 )
 
 const (
@@ -344,4 +345,18 @@ func findHyperShiftMgmtSvcClusters(conn *sdk.Connection, cluster *cmv1.Cluster) 
 
 	// Shouldn't normally happen as every management cluster should have a service cluster
 	return mgmtClusterName, ""
+}
+
+func PrintClusterWarnings(connection *sdk.Connection, cluster *cmv1.Cluster) error {
+	serviceLogs, err := connection.ServiceLogs().V1().Clusters().ClusterLogs().List().ClusterID(cluster.ID()).Send()
+	if err != nil {
+		return err
+	}
+	serviceLogs.Items().Each(func(entry *slv1.LogEntry) bool {
+		if entry.Severity() == slv1.SeverityWarning {
+			fmt.Printf("⚠️ WARNING:\n%s\n%s\n", entry.Summary(), entry.Description())
+		}
+		return true
+	})
+	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/cluster/describe.go
```diff
@@ -345,3 +346,17 @@ func findHyperShiftMgmtSvcClusters(conn *sdk.Connection, cluster *cmv1.Cluster)
...
 	return mgmtClusterName, ""
 }
+
+func PrintClusterWarnings(connection *sdk.Connection, cluster *cmv1.Cluster) error {
+	serviceLogs, err := connection.ServiceLogs().V1().Clusters().ClusterLogs().List().ClusterID(cluster.ID()).Send()
+	if err != nil {
+		return err
+	}
+	serviceLogs.Items().Each(func(entry *slv1.LogEntry) bool {
+		if entry.Severity() == slv1.SeverityWarning {
```
gdbranco, line 356: I think we have other warnings we currently send as service logs. What if the customer would like to not be bothered about it :thinking: . Should this be behind some flag within describe? like --warnings

renan-campos: We do send other warnings as service logs, but none of them come through immediately after installation. I agree that constantly receiving warnings would be a bad user experience. Ultimately it is a BU decision whether warnings get displayed or not. I think they should be displayed by default, with the policy that if the user is seeing the warning, then it describes an action that the user can take to get rid of the warning.  In the one shown here, the user can alleviate it by enabling the Org Policy API.

If it helps push this PR forward, I can add a --quiet flag for users that do not want to see any warnings. We can also filter the service logs to only show warnings of type "Cluster Lifecycle" if needed. Please let me know.

gdbranco (approved):

PR #592, opened 2024-02-01
OCM-1888: Add docs and implement ocm delete account command

cgiradkar:
Prevously when using the help in OCM-CLI it did not list all available sub commands.
This was due to how we currently alias urls in the CLI. Now added a separate sub-command
for account which lets the user delete an account and its associated documentation is also added.

Signed-off-by: Chetan Giradkar <cgiradka@redhat.com>


cmd/ocm/delete/cmd.go
```diff
@@ -40,14 +40,32 @@ var args struct {
...
-	Use:       "delete [flags] PATH",
+	Use:       "delete [flags] PATH|RESOURCE_ALIAS(with resource_id)",
 	Short:     "Send a DELETE request",
 	Long:      "Send a DELETE request to the given path.",
 	RunE:      run,
 	ValidArgs: urls.Resources(),
 }
 
+var arguements_desc_list = map[string]string{
+	"account": `# Delete an account with ACCOUNT_ID
```
mnecas, line 51: Think it would be worth to also document the shorter versions of the arguments

cmd/ocm/delete/cmd.go
```diff
@@ -40,14 +40,32 @@ var args struct {
...
+  ocm delete [--parameter "deleteAssociatedResources=true"] account ACCOUNT_ID`,
+	"subscription": `# Delete a subscription with SUBSCRIPTION_ID
+  ocm delete subscription SUBSCRIPTION_ID`,
+	"role_binding": `# Delete a role binding with ROLE_BINDING_ID
+  ocm delete role_binding ROLE_BINDING_ID`,
+	"sku_rule": `# Delete an SKU role with SKU_ROLE_ID
+  ocm delete sku_rule SKU_ROLE_ID`,
+	"cluster": `# Delete acluster with CLUSTER_ID
+  ocm delete [--parameter "best_effort=true"] [--parameter "deprovision=true"] [parameter "dry_run=true"] cluster CLUSTER_ID`,
```
mnecas, line 60: these `--parameters` are only specific to cluster?

cmd/ocm/delete/cmd.go
```diff
@@ -40,14 +40,32 @@ var args struct {
...
+	Use:       "delete [flags] PATH|RESOURCE_ALIAS(with resource_id)",
 	Short:     "Send a DELETE request",
 	Long:      "Send a DELETE request to the given path.",
 	RunE:      run,
 	ValidArgs: urls.Resources(),
 }
 
+var arguements_desc_list = map[string]string{
+	"account": `# Delete an account with ACCOUNT_ID
+  ocm delete [--parameter "deleteAssociatedResources=true"] account ACCOUNT_ID`,
```
mnecas, line 52: Where is the `deleteAssociatedResources` definied I can't find it in the ocm-cli?

cgiradkar: I was referring to https://api.integration.openshift.com/?urls.primaryName=Accounts%20management%20service#/default/delete_api_accounts_mgmt_v1_accounts__id_

cmd/ocm/delete/cmd.go
cgiradkar: even I have limited information about this one

cmd/ocm/delete/cmd.go
```diff
@@ -40,14 +40,32 @@ var args struct {
...
 var Cmd = &cobra.Command{
-	Use:       "delete [flags] PATH",
+	Use:       "delete [flags] PATH|RESOURCE_ALIAS(with resource_id)",
 	Short:     "Send a DELETE request",
 	Long:      "Send a DELETE request to the given path.",
 	RunE:      run,
 	ValidArgs: urls.Resources(),
 }
 
+var arguements_desc_list = map[string]string{
```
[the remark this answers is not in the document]
cgiradkar: I see no use for map right now but i thought it would be convenient to map it just in case. I'll be changing back to string

cmd/ocm/delete/cmd.go
miguelsorianod: If we wanted to be specific we could document what query parameters are supported for each resource type.

However, looking for example at the `list clusters` CLI command we don't list the specific supported parameters so I am not sure whether it is worth it.

I would like other views around it. Any preference about this @etabak ?

cmd/ocm/delete/cmd.go
```diff
@@ -40,14 +40,49 @@ var args struct {
 }
 
 var Cmd = &cobra.Command{
-	Use:       "delete [flags] PATH",
+	Use:       "delete [flags] PATH|RESOURCE_ALIAS(with resource_id)",
```
miguelsorianod, line 43: I think we should be a bit more aligned with common documentation practices around command lines.

I propose two possible approaches:

One approach is listing the alternative uses:
```
delete [flags] PATH
delete [flags] RESOURCE_ALIAS RESOURCE_ID
```

The other is documenting it in a single line:
```
delete [flags] {PATH|RESOURCE_ALIAS RESOURCE_ID}
```

tzvatot: I would suggest following the convention used by `oc` openshift CLI

miguelsorianod: `oc` / `kubectl` seems to use the single line approach (at least quickly checking some examples in there like `oc label`), with a different character for grouping:

I think it should be in the following way then:
```
delete [flags] (PATH | RESOURCE_ALIAS RESOURCE_ID)
```

cmd/ocm/delete/cmd.go
miguelsorianod: I don't think we need to document every example invocation in the usage. That should belong in the examples section.

In the usage section we should only document the allowed RESOURCE_ALIAS values

cmd/ocm/delete/cmd.go
```diff
@@ -40,14 +40,49 @@ var args struct {
...
 	Short:     "Send a DELETE request",
 	Long:      "Send a DELETE request to the given path.",
 	RunE:      run,
 	ValidArgs: urls.Resources(),
 }
 
+var argumentsUsage = map[string]string{
```
mnecas, line 50: Noticed that when I rerun the help it generated always in a different order which is caused by the map. Either we could implement some ordered map or having just slice of keys in which the map should print the values which don't feel like would be worth for our case or just use the slice with ordered docs.

cgiradkar: Will make it consistent

cmd/ocm/delete/cmd.go
```diff
@@ -40,14 +40,67 @@ var args struct {
...
+
+Use "{{.CommandPath}} [command] --help" for more information about a command.{{end}}
+`
+
+var helpTemplate = `{{with (or .Long .Short)}}{{. | trimTrailingWhitespaces}}
+
+{{end}}{{if or .Runnable .HasSubCommands}}{{.UsageString}}{{end}}`
+
+var argumentsExamples = []string{
+	`  account
```
mnecas, line 86: Why does it need to have additional white space in the beginning?

cmd/ocm/delete/cmd.go
```diff
@@ -40,14 +40,67 @@ var args struct {
...
 var Cmd = &cobra.Command{
-	Use:       "delete [flags] PATH",
+	Use:       "delete [flags] (PATH | RESOURCE_ALIAS RESOURCE_ID)",
 	Short:     "Send a DELETE request",
 	Long:      "Send a DELETE request to the given path.",
 	RunE:      run,
 	ValidArgs: urls.Resources(),
 }
 
+var usageTemplate = `Usage:{{if .Runnable}}
```
mnecas, line 50: If we will adapt this template for other command we should move it to some utils so we don't need to move it when doing refactor of the other commands.

cmd/ocm/delete/cmd.go
```diff
@@ -40,14 +40,67 @@ var args struct {
...
+Global Flags:
+{{.InheritedFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasHelpSubCommands}}
+
+Additional help topics:{{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
+{{rpad .CommandPath .CommandPathPadding}} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableSubCommands}}
+
+Use "{{.CommandPath}} [command] --help" for more information about a command.{{end}}
+`
+
+var helpTemplate = `{{with (or .Long .Short)}}{{. | trimTrailingWhitespaces}}
```
mnecas, line 81: Could you please give a stdout example of these command helps?

cmd/ocm/delete/cmd.go
cgiradkar: to indent all the aliases. 
Currently the o/p looks like this:
image

cmd/ocm/delete/cmd.go
cgiradkar: image

cmd/ocm/delete/cmd.go
cgiradkar: this template was picked up from cobra's internal help/usage template function: https://github.com/spf13/cobra/blob/bcfcff729ecdeb4072ef6f2a9c0b5e4ca1853206/command.go#L546

cmd/ocm/delete/cmd.go
```diff
@@ -40,14 +40,67 @@ var args struct {
 }
 
 var Cmd = &cobra.Command{
-	Use:       "delete [flags] PATH",
+	Use:       "delete [flags] (PATH | RESOURCE_ALIAS RESOURCE_ID)",
```
mnecas, line 43: s/`(PATH | RESOURCE_ALIAS RESOURCE_ID)`/`(PATH|RESOURCE_ALIAS RESOURCE_ID)`/
I would remove the spaces to make it more transparent that it's PATH or RESOURCE_ALICAS.

miguelsorianod: This format was taken from how `kubectl`/`oc` prints it. For example:

```
kubectl label --help
...
Usage:
  kubectl label [--overwrite] (-f FILENAME | TYPE NAME) KEY_1=VAL_1 ... KEY_N=VAL_N [--resource-version=version]
[options]
...
``

cmd/ocm/delete/cmd.go
cgiradkar: removed additional spacing

cgiradkar:
Studied the template format from: https://pkg.go.dev/text/template to make necessary changes. Final o/p:
`$ocm delete -h`
image

cgiradkar:
Added a 2nd commit with format in string literal. Both commits to be compared side by side. @miguelsorianod , can you please compare and lets decide on what's the best format to merge?
I can then remove one of the commits before merging.

miguelsorianod:
If the different parts of the template are not reused anywhere else I am fine with having it as a literal

cmd/ocm/delete/cmd.go
miguelsorianod: +1 about the indentation

cgiradkar:
As discussed, I have used string literal format and added back the 2nd usage (`[command]`)  and the O/P looks like following:
image

cmd/ocm/delete/cmd.go
cgiradkar: Addressed in latest commit

mnecas (approved):
@miguelsorianod could you ptal one last time?

miguelsorianod:
It looks good to me.

I am assuming the output looks like the one shown in https://github.com/openshift-online/ocm-cli/pull/592#issuecomment-1997791438.

If that's not the case please upload a new capture to show the result.

miguelsorianod (approved):

PR #599, opened 2024-02-12
OCM-3226 | feat: improving ocm login and ocm list rh-region url resolution

macgregor:
This MR tweaks the `ocm login --url` resolution algorithm to try to reuse `cfg.URL` before falling back to "api.openshift.com". New `--url` resolution algorithm (from highest priority to lowest priority):

1. runtime `--url` cli arg (key found in `urlAliases`)
2. runtime `--url` cli arg (non-empty string)
3. [NEW] config file `URL` value (non-empty string)
4. sdk.DefaultURL

I also improved the recently added `ocm list rh-region` to be a bit more helpful:
* show discovery URL used
* add optional `--discovery-url` flag with the same url resolution behavior as `ocm login`
* tabular formatting
* display region name and its associated url rather than just region name

### DIFF
--- a/cmd/ocm/list/rhRegion/cmd.go
+++ b/cmd/ocm/list/rhRegion/cmd.go
@@ -2,11 +2,20 @@ package rhRegion
 
 import (
 	"fmt"
+	"os"
+	"strings"
+	"text/tabwriter"
+
 	sdk "github.com/openshift-online/ocm-sdk-go"
 
 	"github.com/openshift-online/ocm-cli/pkg/config"
+	"github.com/openshift-online/ocm-cli/pkg/urls"
 	"github.com/spf13/cobra"
 )
+
+var args struct {
+	discoveryURL string
+}
 
 var Cmd = &cobra.Command{
 	Use:   "rh-regions",
@@ -18,24 +27,40 @@ ocm list rh-regions`,
 	Hidden: true,
 }
 
+func init() {
+	flags := Cmd.Flags()
+	flags.StringVar(
+		&args.discoveryURL,
+		"discovery-url",
+		"",
+		"URL of the OCM API gateway. If not provided, will reuse the URL from the configuration "+
+			"file or "+sdk.DefaultURL+" as a last resort. The value should be a complete URL "+
+			"or a valid URL alias: "+strings.Join(urls.ValidOCMUrlAliases(), ", "),
+	)
+}
+
 func run(cmd *cobra.Command, argv []string) error {
-	// Load the configuration file:
-	cfg, err := config.Load()
+
+	cfg, _ := config.Load()
+
+	gatewayURL, err := urls.ResolveGatewayURL(args.discoveryURL, cfg)
 	if err != nil {
-		return fmt.Errorf("Can't load config file: %v", err)
-	}
-	if cfg == nil {
-		return fmt.Errorf("Not logged in, run the 'login' command")
+		return err
 	}
 
-	regions, err := sdk.GetRhRegions(cfg.URL)
+	fmt.Fprintf(os.Stdout, "Discovery URL: %s\n\n", gatewayURL)
+	regions, err := sdk.GetRhRegions(gatewayURL)
 	if err != nil {
 		return fmt.Errorf("Failed to get OCM regions: %w", err)
 	}
 
-	for regionName := range regions {
-		fmt.Println(regionName)
+	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', tabwriter.TabIndent)
+	fmt.Fprintf(writer, "RH Region\t\tGateway URL\n")
+	for regionName, region := range regions {
+		fmt.Fprintf(writer, "%s\t\t%v\n", regionName, region.URL)
 	}
-	return nil
+
+	err = writer.Flush()
+	return err
 
 }
--- a/cmd/ocm/login/cmd.go
+++ b/cmd/ocm/login/cmd.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/openshift-online/ocm-cli/pkg/config"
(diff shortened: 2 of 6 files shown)


macgregor:
New url config fallback behavior:
```
# no change: --url gets saved to config
> ocm login --token=$OCM_OFFLINE_ACCESS_TOKEN --url=int && ocm config get url
https://api.integration.openshift.com

# previous: when no --url is provided, this would log in against api.openshift.com
# new behavior: fallback to cfg.URL if its available
> ocm login --token=$OCM_OFFLINE_ACCESS_TOKEN && ocm config get url 
https://api.integration.openshift.com
```

New url fallback behavior plus `--rh-region` override behavior:
```
# no change: --url gets saved
> ocm login --token=$OCM_OFFLINE_ACCESS_TOKEN --url=int && ocm config get url
https://api.integration.openshift.com

# no change: --url used for discovery, --rh-region url gets saved to cfg.URL 
> ocm login --token=$OCM_OFFLINE_ACCESS_TOKEN --url=stage --rh-region=rh-singapore && ocm config get url
https://api.aws.xcm.stage.openshift.com

# previous: when no --url provided, api.openshift.com would be used for service discovery when --rh-region is provided
# new behavior: use cfg.URL if its available for service discovery when --rh-region is provided 
> ocm login --token=$OCM_OFFLINE_ACCESS_TOKEN --rh-region=rh-singapore && ocm config get url 
https://api.aws.xcm.stage.openshift.com
```

`ocm list rh-regions` improvements:
```
# no change: cfg.URL is used for service discovery
# new behaviour: print discovery url used, table format, print discovered gateway urls rather than just region names
> ocm login --token=$OCM_OFFLINE_ACCESS_TOKEN --url=stage && ocm list rh-regions                  
Discovery URL: https://api.stage.openshift.com

RH Region       Gateway URL
rh-na           api.stage.openshift.com
rh-singapore    api.aws.xcm.stage.openshift.com

> ocm login --token=$OCM_OFFLINE_ACCESS_TOKEN --url=stage && ocm list rh-regions --discovery-url=int
Discovery URL: https://api.integration.openshift.com

RH Region       Gateway URL
rh-na           api.integration.openshift.com
rh-singapore    api.aws.xcm.integration.openshift.com
```

tylercreller (approved):

tirthct:
Can we add some tests here to confirm the behavior?

cmd/ocm/list/rhRegion/cmd.go
```diff
@@ -18,24 +27,51 @@ ocm list rh-regions`,
...
-	regions, err := sdk.GetRhRegions(cfg.URL)
+	fmt.Fprintf(os.Stdout, "Discovery URL: %s\n\n", gatewayURL)
+	regions, err := sdk.GetRhRegions(gatewayURL)
 	if err != nil {
 		return fmt.Errorf("Failed to get OCM regions: %w", err)
 	}
 
-	for regionName := range regions {
-		fmt.Println(regionName)
+	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', tabwriter.TabIndent)
```
tirthct, line 68: Can we use dump.Pretty() here? https://github.com/openshift-online/ocm-cli/blob/main/pkg/dump/dump.go#L33

cmd/ocm/login/cmd.go
```diff
@@ -295,11 +278,39 @@ func run(cmd *cobra.Command, argv []string) error {
...
-	// If the value of the `--url` is any of the aliases then replace it with the corresponding
-	// real URL:
-	gatewayURL, ok := urlAliases[args.url]
-	if !ok {
+	// URL Precedent (from highest priority to lowest priority):
+	//   1. runtime `--url` cli arg (key found in `urlAliases`)
+	//   2. runtime `--url` cli arg (non-empty string)
+	//   3. config file `URL` value (non-empty string)
+	//   4. sdk.DefaultURL
+	gatewayURL := sdk.DefaultURL
```
tirthct, line 286: Do you think we should add this to a function? We are re-using this in list rh-region command as well

cmd/ocm/list/rhRegion/cmd.go
macgregor: Im not writing the raw json, which is what dump seems to be mostly for. Theres a pretty even mix of usage between dump and tabwriter in the repo. In this case, I was copying the pattern from https://github.com/openshift-online/ocm-cli/blob/main/cmd/ocm/list/region/cmd.go#L106 which seemed like the most similar command. 

A more sophisticated cli might support variable output, like `oc get ... -o json` so we could easily pick and choose which format we wanted, but that doesnt seem to be a first class concept in ocm at the moment. For now ill leave it as the `tabwriter` unless someone else comes along with a stronger opinion.

macgregor:
> Can we add some tests here to confirm the behavior?

added tests for the now shared `ResolveGatewayURL` method

pkg/urls/well_known.go
```diff
@@ -18,5 +18,54 @@ limitations under the License.
...
+	}
+	return keys
+}
+
+// URL Precedent (from highest priority to lowest priority):
+//  1. runtime `--url` cli arg (key found in `urlAliases`)
+//  2. runtime `--url` cli arg (non-empty string)
+//  3. config file `URL` value (non-empty string)
+//  4. sdk.DefaultURL
+func ResolveGatewayURL(optionalParsedCliFlagValue string, optionalParsedConfig *config.Config) string {
```
tirthct, line 59: How do we handle a scenario where:
User wants to log in to prod : `ocm login --url=Prod`
But this would fail since Prod is not a valid alias.
We have a stage URL stored in config, so we log the user into OCM stage.
Should we show any proper messages in this case?

macgregor: I dont follow, "production", "prod" and "prd" are all valid aliases. If you mean how do we detect a mistaken prod alias vs a raw url and give a better error message, we dont do that today. I can add that, though.

macgregor: So its tricky to know the user's exact intent. Were they trying to use an alias and typoed it or used one that doesnt exist, or did they input an invalid url. Which is probably why they didnt bother before, they just pass the this-is-not-a-known-alias `--url` value to the underlying SDK and let it throw up an error if it wasnt actually a good url. 

So I made ResolveGatewayURL, as a final step, try to parse the resolved URL using url.ParseRequestURI and return an helpful error message if that fails.

```
> ocm login --token=$OCM_OFFLINE_ACCESS_TOKEN --url="Prd"      
Error: parse "Prd": invalid URI for request

URL Source: flag
Expected an absolute URI or an absolute path (e.g. https://api.openshift.com) or a case-sensitive alias, one of: [int, production, prod, prd, staging, stage, stg, integration]
```

Logic interpreted that as the user trying to input an invalid url because its not a known alias, but we display our expectations, including the known aliases and the fact that they are case sensitive in the error message to help the user know what to do about it. As a bonus, this also works if their saved `cfg.URL` gets mangled for some reason.

macgregor: Also to be clear, even before i improved things `ocm login --url=Prod` would have failed, not fall back to the stored config, just like it does on main today. It would have just given you a slightly less helpful error message

ciaranRoche (approved):
lgtm

PR #601, opened 2024-02-28
OCM-6407 | edit option sends an empty payload

ahitacat:
## EDIT
This commit adds a parameter when creating the command that check for a minimum amount of arguments. The
command will fail if no parameters are used.

### DIFF
--- a/cmd/ocm/edit/cluster/cmd.go
+++ b/cmd/ocm/edit/cluster/cmd.go
@@ -47,6 +47,7 @@ var Cmd = &cobra.Command{
 	Example: `  # Edit a cluster named "mycluster" to make it private
   ocm edit cluster mycluster --private`,
 	RunE: run,
+	Args: cobra.MinimumNArgs(1),
 }
 
 func init() {
@@ -132,7 +133,7 @@ func wasClusterWideProxyReceived(httpProxy, httpsProxy, noProxy, additionalTrust
 }
 
 func run(cmd *cobra.Command, argv []string) error {
-	// Check that there is exactly one cluster name, identifir or external identifier in the
+	// Check that there is exactly one cluster name, identifier or external identifier in the
 	// command line arguments:
 	if len(argv) != 1 {
 		return fmt.Errorf(


tzvatot:
I suggest to use `cmd.Flags().NFlag()` and if it's < 1 then print an INFO message that no change is done.

ahitacat:
I see there is also used in this cli tool the Args check.
```
var Cmd = &cobra.Command{
	...
	Args: cobra.MinimumNArgs(2),
}
```
If I remove the clusterNeedsUpdate bool, I see a possible problem when the --enable-delete-protection parameter is added. As that parameter doesn't require to update the cluster.

tzvatot:
> I see there is also used in this cli tool the [builtin args validator](https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#positional-and-custom-arguments).
> 
> ```
> var Cmd = &cobra.Command{
> 	...
> 	Args: cobra.MinimumNArgs(2),
> }
> ```
> 
> If I remove the clusterNeedsUpdate bool, I see a possible problem when the --enable-delete-protection parameter is added. As that parameter doesn't require to update the cluster.

Why not? it requires a patch to the /delete_protection endpoint. It's not the cluster's endpoint, but it's still semantically updating the cluster.

mnecas:
Agree with @ahitacat in case we will use the `--enable-delete-protection` without adding any check to [1] we will still send empty patch request to the `/clusters/{id}` endpoint and patch to update the delete-protection to `/clusters/{id}/delete_protection`. This could lead to unexpected behaviour as we would not update anything in the  `/clusters/{id}`  but still could fail on permission check. 

[1] https://github.com/openshift-online/ocm-cli/blob/109526466d14f5daf8f169cba211e672c0ea4f6e/cmd/ocm/edit/cluster/cmd.go#L271

mnecas:
Ah did not know about the https://github.com/openshift-online/ocm-cli/pull/603 should be good.
lgmt

tzvatot (approved):
LGTM

PR #603, opened 2024-03-01
OCM-6450 | No update cluster with empty config

ahitacat:
ocm edit will update always the cluster even there is no new configuration. This commit checks the configuration before make the update call.


tzvatot (approved):
LGTM

PR #605, opened 2024-03-04
OCM-6140 | feat: allow customization of the domain prefix when creating a cluster and prints it out in the describe cluster command

machi1990:


cmd/ocm/create/cluster/cmd.go
```diff
@@ -174,6 +176,18 @@ func init() {
 	arguments.SetQuestion(fs, "version", "OpenShift version:")
 	Cmd.RegisterFlagCompletionFunc("version", arguments.MakeCompleteFunc(getVersionOptions))
 
+	fs.StringVar(
+		&args.domainPrefix,
+		"domain-prefix",
+		"",
+		"An optional unique domain prefix of the cluster. This will be used when generating a "+
+			"sub-domain for your cluster on openshiftapps.com. It must be unique and consist of "+
```
tzvatot, line 184: If we ever change the URL from openshiftapps.com to something else, it means every customer that is using an older version of the CLI would see incorrect help message. That said, I don't have a workaround/alternative other than maybe writing something vague w/o specifying the exact domain. 
Anyhow, not a strong objection from writing it, just raising the awareness.

machi1990: That's a good point; I basically went with the help message that was in the `.name`.

machi1990: I could drop the `openshiftapps.com` from the text all together.

machi1990: done

machi1990:
@tzvatot Please give this another round of reviews. Thanks

cmd/ocm/create/cluster/cmd.go
```diff
@@ -174,6 +176,18 @@ func init() {
 	arguments.SetQuestion(fs, "version", "OpenShift version:")
 	Cmd.RegisterFlagCompletionFunc("version", arguments.MakeCompleteFunc(getVersionOptions))
 
+	fs.StringVar(
+		&args.domainPrefix,
+		"domain-prefix",
+		"",
+		"An optional unique domain prefix of the cluster. This will be used when generating a "+
```
tzvatot, line 183: Would be nice to know what will be the behaviour  (default) if user does not provide it. WDYT?

machi1990: Sounds good.

machi1990: done

tzvatot (approved):
LGTM

PR #606, opened 2024-03-07
OCM-6030 | feat: allow to edit component routes of ingress

gdbranco:
OCM-6030

### DIFF
--- a/cmd/ocm/edit/ingress/cmd.go
+++ b/cmd/ocm/edit/ingress/cmd.go
@@ -20,6 +20,7 @@ import (
 
 	c "github.com/openshift-online/ocm-cli/pkg/cluster"
 	"github.com/openshift-online/ocm-cli/pkg/ocm"
+	"github.com/openshift-online/ocm-cli/pkg/utils"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 )
@@ -31,6 +32,15 @@ var ValidWildcardPolicies = []string{string(cmv1.WildcardPolicyWildcardsDisallow
 	string(cmv1.WildcardPolicyWildcardsAllowed)}
 var ValidNamespaceOwnershipPolicies = []string{string(cmv1.NamespaceOwnershipPolicyStrict),
 	string(cmv1.NamespaceOwnershipPolicyInterNamespaceAllowed)}
+var expectedComponentRoutes = []string{
+	string(cmv1.ComponentRouteTypeOauth),
+	string(cmv1.ComponentRouteTypeConsole),
+	string(cmv1.ComponentRouteTypeDownloads),
+}
+var expectedParameters = []string{
+	hostnameParameter,
+	tlsSecretRefParameter,
+}
 
 var args struct {
 	clusterKey    string
@@ -43,6 +53,8 @@ var args struct {
 	namespaceOwnershipPolicy  string
 	clusterRoutesHostname     string
 	clusterRoutesTlsSecretRef string
+
+	componentRoutes string
 }
 
 const (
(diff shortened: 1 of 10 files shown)


cmd/ocm/edit/ingress/cmd.go
```diff
@@ -55,6 +67,12 @@ const (
 	namespaceOwnershipPolicyFlag  = "namespace-ownership-policy"
 	clusterRoutesHostnameFlag     = "cluster-routes-hostname"
 	clusterRoutesTlsSecretRefFlag = "cluster-routes-tls-secret-ref"
+	componentRoutesFlag           = "component-routes"
+
+	amountOfParametersPerComponentRoute = 2
+	hostnameParameter                   = "hostname"
```
cristianoveiga, line 73: nit: indenting seems off

cmd/ocm/edit/ingress/cmd.go
```diff
@@ -55,6 +67,12 @@ const (
...
+	//nolint:gosec
+	tlsSecretRefParameter = "tlsSecretRef"
```
cristianoveiga, line 75: nit: indenting seems off

cmd/ocm/edit/ingress/cmd.go
```diff
@@ -148,6 +166,15 @@ func init() {
 		"",
 		"Components route TLS secret reference for oauth, console, download.",
 	)
+
+	flags.StringVar(
+		&args.componentRoutes,
+		componentRoutesFlag,
+		"",
+		//nolint:lll
+		"Component routes settings. Available keys [oauth, console, download]. For each key a pair of hostname and tlsSecretRef is expected to be supplied. "+
```
cristianoveiga, line 175: should this be `downloads` instead?

cmd/ocm/edit/ingress/cmd.go
gdbranco: Oh it's because of the nolint comment, won't be able to resolve it

cmd/ocm/edit/ingress/cmd.go
gdbranco: This one is just the separation from the parameters above which are all flags

cmd/ocm/edit/ingress/cmd.go
```diff
@@ -55,6 +67,12 @@ const (
 	namespaceOwnershipPolicyFlag  = "namespace-ownership-policy"
 	clusterRoutesHostnameFlag     = "cluster-routes-hostname"
 	clusterRoutesTlsSecretRefFlag = "cluster-routes-tls-secret-ref"
+	componentRoutesFlag           = "component-routes"
+
+	amountOfParametersPerComponentRoute = 2
```
cristianoveiga, line 72: perhaps we could avoid this by adding `hostnameParameter `  and `tlsSecretRefParameter` to its own slice

cristianoveiga: actually you are already doing this in `expectedParameters` - can we just use that instead?

gdbranco: Oh true, I had missed that. Let's adjust!

gdbranco: I'll still need const though, it just so happened that the parameters currently match the [parsed component length](https://github.com/openshift-online/ocm-cli/pull/606/files#diff-e49b4de7ddb1004c3a10c0f5c3867e9c133741d618742f7f14462b1173497232R357), but I would rather keep them separate for future proofing. But I went ahead and renamed for better intention and used the [expectedParams](https://github.com/openshift-online/ocm-cli/pull/606/files#diff-e49b4de7ddb1004c3a10c0f5c3867e9c133741d618742f7f14462b1173497232R373) for the actual param length

cristianoveiga:
LGTM

PR #607, opened 2024-03-08
OCM-6140: make name width to be 54 chars to fix truncation issue in case of longer name > 28 chars

machi1990:
This follows up on https://github.com/openshift-online/ocm-cli/pull/605

### DIFF
--- a/pkg/output/table_test.go
+++ b/pkg/output/table_test.go
@@ -261,8 +261,8 @@ var _ = Describe("Table", func() {
 		// Check the generated text:
 		lines := strings.Split(buffer.String(), "\n")
 		Expect(lines).To(HaveLen(3))
-		Expect(lines[0]).To(Equal(`ID                                NAME                        `))
-		Expect(lines[1]).To(Equal(`123                               mycluster                   `))
+		Expect(lines[0]).To(Equal(`ID                                NAME                                                  `))
+		Expect(lines[1]).To(Equal(`123                               mycluster                                             `))
 	})
 
 	It("Honours learning limit", func() {
--- a/pkg/output/tables/clusters.yaml
+++ b/pkg/output/tables/clusters.yaml
@@ -20,7 +20,7 @@ columns:
   width: 32
 - name: name
   header: NAME
-  width: 28
+  width: 54
 - name: api.url
   header: API URL
   width: 58


machi1990:
@AkashKanni can you give this PR a look, it aims to fix the UX issue you observed? Thanks

pkg/output/table_test.go
```diff
@@ -261,8 +261,8 @@ var _ = Describe("Table", func() {
 		// Check the generated text:
 		lines := strings.Split(buffer.String(), "\n")
 		Expect(lines).To(HaveLen(3))
-		Expect(lines[0]).To(Equal(`ID                                NAME                        `))
-		Expect(lines[1]).To(Equal(`123                               mycluster                   `))
+		Expect(lines[0]).To(Equal(`ID                                NAME                                                  `))
+		Expect(lines[1]).To(Equal(`123                               mycluster                                             `))
```
mnecas, line 265: What is the change here, just whitespace?

mnecas: ahh the GitHub diff did not show the trailing whitespaces
Could we use some string repeat of space?

machi1990: To reflect the output of the table after the increase in width to 54 chars.

machi1990: We could use string repeat of space but I don’t see how that will improve it; I mean visually looking at the output like this this helps because it matches what really is display when the list cmd is run something that we won’t have with string repeat.

pkg/output/tables/clusters.yaml
```diff
@@ -20,7 +20,7 @@ columns:
   width: 32
 - name: name
   header: NAME
-  width: 28
+  width: 54
```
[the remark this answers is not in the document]
AkashKanni: LGTM

pkg/output/table_test.go
mnecas: I was mainly thinking about easy updating of the tests as I don't want to count the exact number of whitespace

or additionally, the issue I had with GitHub ignoring trailing whitespace 
image

pkg/output/tables/clusters.yaml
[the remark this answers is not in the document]
machi1990: Thanks @AkashKanni

pkg/output/table_test.go
ziccardi: I would prefer to keep it like this for readability: with just a quick look it is clear what we are expecting to be inside the `lines`

@mnecas to me github shows clearly the trailing whitespace:
image

mnecas: okay, lgmt

machi1990: Thanks for clarifying your point, which I see. I think it's a tradeoff between visually seeing the expected output vs speed of test update. I personally prefer the former for expected output like this that needs visual inspection.

PR #610, opened 2024-04-30
OCM-6528 | feat: add describe ingress cmd

gdbranco:
OCM-6528

# How
```
ocm describe ingress xxx -i apps
Cluster ID:                 xxx
Default:                    true
ID:                         xxxx
LB-Type:                    nlb
Namespace Ownership Policy: Strict
Private:                    false
Wildcard Policy:            WildcardsDisallowed
```


ciaranRoche (approved):
/lgtm

PR #621, opened 2024-06-04
Adds ability to change api url via env var

iamkirkbater:
Adds the ability to change the OCM API URL based on the environment variable `OCM_URL` allowing a user to change the OCM context without changing global state on their machine.

Satisfies OCM-8459.

# Testing

To test this, I built the binary locally and ran a few commands, expecting different output.

* ./ocm list clusters --parameter search="name like 'hs-mc%'"
* OCM_URL=https://api.[regional-api].openshift.com ./ocm list clusters

### DIFF
--- a/cmd/ocm/account/status/cmd.go
+++ b/cmd/ocm/account/status/cmd.go
@@ -18,11 +18,13 @@ package status
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 
 	acc_util "github.com/openshift-online/ocm-cli/pkg/account"
 	"github.com/openshift-online/ocm-cli/pkg/config"
+	"github.com/openshift-online/ocm-cli/pkg/ocm"
 	amv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 )
 
@@ -61,19 +63,10 @@ func run(cmd *cobra.Command, argv []string) error {
 		return fmt.Errorf("Not logged in, run the 'login' command")
 	}
 
-	// Check that the configuration has credentials or tokens that haven't have expired:
-	armed, reason, err := cfg.Armed()
+	// Create the client for the OCM API:
+	connection, err := ocm.NewConnection().Build()
 	if err != nil {
 		return err
-	}
-	if !armed {
-		return fmt.Errorf("Not logged in, %s, run the 'login' command", reason)
-	}
-
-	// Create the connection, and remember to close it:
-	connection, err := cfg.Connection()
-	if err != nil {
-		return fmt.Errorf("Can't create connection: %v", err)
 	}
 	defer connection.Close()
 
@@ -87,8 +80,12 @@ func run(cmd *cobra.Command, argv []string) error {
 	// Display user and which server they are logged into
 	currAccount := response.Body()
 	currOrg := currAccount.Organization()
-	fmt.Printf("User %s on %s in org '%s' %s (external_id: %s)",
+	fmt.Printf("User %s on %s in org '%s' %s (external_id: %s) ",
 		currAccount.Username(), cfg.URL, currOrg.Name(), currOrg.ID(), currOrg.ExternalID())
+
+	if urlOverride := os.Getenv("OCM_URL"); urlOverride != "" {
+		fmt.Printf("- OCM_URL overridden via env to %s - ", urlOverride)
+	}
 
 	// Display roles currently assigned to the user
 	roleSlice, err := acc_util.GetRolesFromUsers([]*amv1.Account{currAccount}, connection)
--- a/cmd/ocm/account/status/cmd.go
+++ b/cmd/ocm/account/status/cmd.go
@@ -18,7 +18,6 @@ package status
 
 import (
 	"fmt"
-	"os"
 
 	"github.com/spf13/cobra"
 
@@ -82,10 +81,6 @@ func run(cmd *cobra.Command, argv []string) error {
 	currOrg := currAccount.Organization()
 	fmt.Printf("User %s on %s in org '%s' %s (external_id: %s) ",
 		currAccount.Username(), cfg.URL, currOrg.Name(), currOrg.ID(), currOrg.ExternalID())
-
-	if urlOverride := os.Getenv("OCM_URL"); urlOverride != "" {
-		fmt.Printf("- OCM_URL overridden via env to %s - ", urlOverride)
-	}
 
 	// Display roles currently assigned to the user
 	roleSlice, err := acc_util.GetRolesFromUsers([]*amv1.Account{currAccount}, connection)
--- a/cmd/ocm/login/cmd.go
+++ b/cmd/ocm/login/cmd.go
@@ -315,7 +315,7 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 
 	if overrideUrl := os.Getenv(properties.URLEnvKey); overrideUrl != "" {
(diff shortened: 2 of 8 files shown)


cmd/ocm/account/status/cmd.go
```diff
@@ -87,9 +80,13 @@ func run(cmd *cobra.Command, argv []string) error {
 	// Display user and which server they are logged into
 	currAccount := response.Body()
 	currOrg := currAccount.Organization()
-	fmt.Printf("User %s on %s in org '%s' %s (external_id: %s)",
+	fmt.Printf("User %s on %s in org '%s' %s (external_id: %s) ",
 		currAccount.Username(), cfg.URL, currOrg.Name(), currOrg.ID(), currOrg.ExternalID())
 
+	if urlOverride := os.Getenv("OCM_URL"); urlOverride != "" {
+		fmt.Printf("- OCM_URL overridden via env to %s - ", urlOverride)
+	}
```
iamkirkbater, line 88: Changes the output based on whether the env is overridden:

```
> ./ocm account status
User testuser on https://api.openshift.com in org 'Red Hat' aaabbbcccdddeeefffggg (external_id: 123456789) Roles: Role1, Role2, AuthenticatedUser

ocm-cli on  api-url-from-env [!]  staging
> OCM_URL=https://regional.api.openshift.com ./ocm account status
User testuser on https://api.openshift.com in org 'Red Hat' aaabbbcccdddeeefffggg (external_id: 123456789) - OCM_URL overridden via env to https://regional.api.openshift.com - Roles: Role1, Role2, AuthenticatedUser```

Comment (iamkirkbater) [reply cmd/ocm/account/status/cmd.go:88]: Happy to handle this differently if there's any better ideas of how to format this.

File: pkg/ocm/connection.go [hunk 2]
```diff
@@ -63,6 +65,16 @@ func (b *ConnectionBuilder) Build() (result *sdk.Connection, err error) {
...
+	// overwrite the config URL if the environment variable is set
+	if overrideUrl := os.Getenv("OCM_URL"); overrideUrl != "" {
+		if debug.Enabled() {
+			fmt.Fprintln(os.Stderr, "INFO: OCM_URL is overridden via environment variable. This functionality is considered tech preview and may cause unexpected issues.")
+			fmt.Fprintln(os.Stderr, "      If you experience issues while OCM_URL is set, unset the OCM_URL environment variable and attempt to log in directly to the desired OCM environment.")
+			fmt.Fprintln(os.Stderr, "") // add a newline on purpose to separate from ocm sdk debug output
+		}
+		b.cfg.URL = overrideUrl
+	}
+
```
iamkirkbater, line 77: When run with the debug flag, output will end up looking like this:
```
> OCM_URL=https://api.regional.openshift.com ./ocm --debug get cluster $CLUSTER_ID
INFO: OCM_URL is overridden via environment variable. This functionality is considered tech preview and may cause unexpected issues.
      If you experience issues while OCM_URL is set, unset the OCM_URL environment variable and attempt to log in directly to the desired OCM environment.
      
I0604 10:23:47.595646   83670 connection.go:876] Added URL with prefix '', regular expression '^(/.*)?$' and URL 'https://api.regional.openshift.com'
...
```

cmd/ocm/login/cmd.go
```diff
@@ -313,6 +313,10 @@ func run(cmd *cobra.Command, argv []string) error {
 		gatewayURL = fmt.Sprintf("https://%s", regValue.URL)
 	}
 
+	if overrideUrl := os.Getenv("OCM_URL"); overrideUrl != "" {
+		fmt.Fprintf(os.Stderr, "WARNING: the `OCM_URL` environment variable is set, but is not used for the login command. The `ocm login` command will only use the explicitly set flag's url, which is set as %s\n", gatewayURL)
+	}
+
```
iamkirkbater, line 319:
```
> OCM_URL=https://api.stage.openshift.com ./ocm login --token=$OCM_ACCESS_TOKEN --url integration
WARNING: the `OCM_URL` environment variable is set, but is not used for the login command. The `ocm login` command will only use the explicitly set flag's url, which is set as https://api.integration.openshift.com

> ocm config get url
https://api.integration.openshift.com
```

pkg/ocm/connection.go
```diff
@@ -63,6 +65,16 @@ func (b *ConnectionBuilder) Build() (result *sdk.Connection, err error) {
...
 	}
 
+	// overwrite the config URL if the environment variable is set
+	if overrideUrl := os.Getenv("OCM_URL"); overrideUrl != "" {
+		if debug.Enabled() {
+			fmt.Fprintln(os.Stderr, "INFO: OCM_URL is overridden via environment variable. This functionality is considered tech preview and may cause unexpected issues.")
+			fmt.Fprintln(os.Stderr, "      If you experience issues while OCM_URL is set, unset the OCM_URL environment variable and attempt to log in directly to the desired OCM environment.")
+			fmt.Fprintln(os.Stderr, "") // add a newline on purpose to separate from ocm sdk debug output
+		}
+		b.cfg.URL = overrideUrl
```
macgregor, line 75: You should check for references to ` config.Save()` like https://github.com/openshift-online/ocm-cli/blob/main/cmd/ocm/get/cmd.go#L126

I believe any command that calls save after `ocm.NewConnection().Build()` will result in that env override being persisted to the config json. Which might be ok in most situations, but is unexpected and best avoided.

cmd/ocm/account/status/cmd.go
macgregor: imo just nix it. your debug warning when you override the url in ConnectionBuilder covers all use cases and doesnt create noise.

pkg/ocm/connection.go
iamkirkbater: I'll double check the other commands for this one - but in this case the `cfg` is loaded and then `connection` is built by loading the cfg again. So at least in this case, since we're not getting the `cfg` back from the connection (it's only returning the updated tokens) the URL is not being passed back and is not being updated in the original `cfg` var.

With that said, though, I'll explicitly double check for other instances to make sure we're not inadvertently saving the cfg back from the connection itself and then overwriting the config.

iamkirkbater: Double checked all instances of `config.Save()` that were found and verified that each of them are loading the config separately of the ConnectionBuilder struct, so any changes within ConnectionBuilder would not affect the saving of the refresh/access tokens.

With that said, it probably wouldn't be a terrible idea to further refactor the refresh/access token logic since it's pretty duplicated throughout the codebase, but I would consider that out of scope for this PR specifically.

cmd/ocm/login/cmd.go
```diff
@@ -313,6 +313,10 @@ func run(cmd *cobra.Command, argv []string) error {
 		gatewayURL = fmt.Sprintf("https://%s", regValue.URL)
 	}
 
+	if overrideUrl := os.Getenv("OCM_URL"); overrideUrl != "" {
```
tylercreller, line 316: I've recently created this properties file we can throw env variable in https://github.com/openshift-online/ocm-cli/blob/main/pkg/properties/properties.go

iamkirkbater: Nice - updated :D

macgregor (approved):
lgtm

would be nice if we had designed env overrides in the first place, but it think this is fine as a one off specifically for url. especially in the context of regional ocm where switching urls will be more common than just stage/prod today.

iamkirkbater:
Yeah - if you want to rewrite the whole thing I'd consider implementing `viper`, however the login/config file management used for auth token handling may not work with viper out of the gate and would then end up being just as much of a headache.

tylercreller (approved):

cristianoveiga (approved):

renan-campos (approved):
This cleans up the codebase and provides a convenient way to interact with different environments from the same machine. Thank you for this contribution.

PR #630, opened 2024-07-03
Initial refactor to prepare to move some packages to ocm-common

iamkirkbater:
Some initial refactors in order to prepare to move the `pkg/ocm/connection-builder` (formerly `pkg/ocm`) and `pkg/config` to ocm-common to be reusable between multiple projects.

## Key Features
* Removes the `Connection` method from `pkg/config` - this just felt completely out of place in the config loader logic and felt like it overloaded the config package and was out of scope
* Allows passing a custom logger to the SDK (this is more of a feature for when this moves to ocm-common so that consumers can pass their own logger in so long as it implements the interface expected)
* Allows overriding the URL via a function in order to keep env var checking out of the connection builder itself - move this to the consumer to define (you can see how this would be used in the `cmd/login` function - we explicitly want to override the URL with the same value as the config in the case that the env var is set and the order in which the new `pkg/ocm/connection.go` will check for these things
* Speaking of `pkg/ocm/connection.go` - this file is intended to set the defaults for all of this CLI while not changing the flow that's already implemented in most of the `cmd/*` files.
* Updated the few `cmd/*` files that were still modifying the config directly - some are new from when we merged the PR but the login command also needed to be changed to work with the new model.
* ConnectionBuilder Allows overriding the Agent string - also more important for when this moves to ocm-common so that downstream consumers of ocm-common will be able to define their own agent string - so you'll know if any given request is coming from external tooling rather than the OCM cli itself.

I'd very much like to add some unit tests to this, and will be going on PTO for the next few weeks, so I will be horribly unresponsive, but I just wanted to get a sneak preview of this in now so that any feedback could be collected.

### DIFF
--- a/cmd/ocm/login/cmd.go
+++ b/cmd/ocm/login/cmd.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/openshift-online/ocm-cli/pkg/config"
+	"github.com/openshift-online/ocm-cli/pkg/ocm"
 	"github.com/openshift-online/ocm-cli/pkg/properties"
 	"github.com/openshift-online/ocm-cli/pkg/urls"
 	sdk "github.com/openshift-online/ocm-sdk-go"
@@ -329,7 +330,7 @@ func run(cmd *cobra.Command, argv []string) error {
 	cfg.Insecure = args.insecure
 
 	// Create a connection and get the token to verify that the crendentials are correct:
-	connection, err := cfg.Connection()
+	connection, err := ocm.NewConnection().Config(cfg).WithApiUrl(gatewayURL).Build()
 	if err != nil {
 		return fmt.Errorf("Can't create connection: %v", err)
 	}
--- a/cmd/ocm/gcp/get-wif-config.go
+++ b/cmd/ocm/gcp/get-wif-config.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/openshift-online/ocm-cli/pkg/config"
 	"github.com/openshift-online/ocm-cli/pkg/dump"
+	"github.com/openshift-online/ocm-cli/pkg/ocm"
 	"github.com/openshift-online/ocm-cli/pkg/urls"
 	"github.com/spf13/cobra"
 )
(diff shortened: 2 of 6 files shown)


ciaranRoche:
Gave this a first pass, and compared to what we have in ROSA CLI, I think this is a good starting point to lift both `pkg/config` and `pkg/ocm` to ocm-common and share between both clients. 

Thanks for putting this together

iamkirkbater:
I didn't get a chance today to write any unit tests like I had hoped, so I am removing this from Draft status and will freely allow you to merge this if you want in order to potentially migrate to ocm-common to be used by other repos before I get back from PTO and don't want to keep this all blocked up for no real good reason.

iamkirkbater:
Looking at the connection-builder package here I'm not sure if it's worth writing tests for, since the logic is relatively basic of just setting and getting values. Let me know if you think otherwise and I'll add them, though.  With that said, I think this is ready for review whenever y'all get a chance :)

cmd/ocm/login/cmd.go
```diff
@@ -329,7 +330,7 @@ func run(cmd *cobra.Command, argv []string) error {
 	cfg.Insecure = args.insecure
 
 	// Create a connection and get the token to verify that the crendentials are correct:
-	connection, err := cfg.Connection()
+	connection, err := ocm.NewConnection().Config(cfg).WithApiUrl(gatewayURL).Build()
```
tylercreller, line 333: WithApiUrl looks redundant with that same URL also being set in the config, is there a particular use case it is needed in this spot?

tylercreller:
Overall LGTM - one remaining comment

cmd/ocm/login/cmd.go
iamkirkbater: In this case the config is loaded but the URL is provided by the flag. 

In a case where someone's config is currently loaded with Staging but they pass in the `--url production` to the login command, this would ensure that the login command's `--url` flag is respected.

tylercreller: Interesting, I would think that with `config.URL` being assigned on line 327 with the same gatewayURL and then used in the builder pattern, that it would be used to build with the updated flag value. Just to confirm, are you saying that will not be the case? I could be reading this wrong.

iamkirkbater: Looking at it again I think that you're actually correct here. Let me make and test the change real quick!

iamkirkbater: Worked just fine, great catch!

tylercreller (approved):

ciaranRoche (approved):
/lgtm

PR #643, opened 2024-07-19
Replace wif models and client with sdk

JakobGray:
This PR will
- Update to the latest version of the ocm-sdk
- Remove old WIF models and custom client code and replace them with the sdk
- Update all reference to the old models in `cmd/ocm/gcp` with the official SDK models and client

It is not intended to introduce changes in functionality; however some changes were necessary to handle custom roles.


JakobGray:
@ckandag @renan-campos

renan-campos (commented):
Please use audience and impersonator email from the sdk.

renan-campos (approved):

JakobGray:
I found issues with the wif templates that need to be addressed before this is merged

cmd/ocm/gcp/delete-wif-config.go
```diff
@@ -59,21 +58,23 @@ func deleteWorkloadIdentityConfigurationCmd(cmd *cobra.Command, argv []string) {
 		log.Fatal("WIF config ID is required")
 	}
 
-	// Create clients
-	ocmClient, err := alphaocm.NewOcmClient()
+	// Create the client for the OCM API:
+	connection, err := ocm.NewConnection().Build()
 	if err != nil {
-		log.Fatalf("failed to create backend client: %v", err)
+		log.Fatal(err)
```
ckandag, line 64: isnt this missing a return stmt when error occurs ?

ckandag: I thought previously discussed bubbling up the errors to the caller and allow the top level "Run" funtion decide how to handle/display the error.

I dont see these kind of Fatal log msgs anywhere in the else in the ocm or rosa cli code

ckandag:
Across the board please fix error handling to be similar to be rest of ocm cli.
Please remove all the Fatal stmt and bubble errors up to caller. This allows for the error handling/ display  to be addressed better and consistently across the CLI

cmd/ocm/gcp/delete-wif-config.go
JakobGray: An example error:
```
➜  go run ./cmd/ocm gcp delete wif-config 2crdacmjbo3jmdgfn4dq791a7nig04qg
2024/07/30 14:25:07 Not logged in, credentials aren't set, run the 'login' command
exit status 1
```
This is the top level Run function.

JakobGray:
> @JakobGray : Across the board please fix error handling to be similar to be rest of ocm cli. Please remove all the Fatal stmt and bubble errors up to caller. This allows for the error handling/ display to be addressed better and consistently across the CLI

Because this PR is not intended to introduce changes in functionality I don't want to extend it's scope to address all error handling. I believe we have a ticket already to address error handling more gracefully so I think that kind of change is better addressed there.

ckandag:
> > @JakobGray : Across the board please fix error handling to be similar to be rest of ocm cli. Please remove all the Fatal stmt and bubble errors up to caller. This allows for the error handling/ display to be addressed better and consistently across the CLI
> 
> Because this PR is not intended to introduce changes in functionality I don't want to extend it's scope to address all error handling. I believe we have a ticket already to address error handling more gracefully so I think that kind of change is better addressed there.

@JakobGray  I raised this in the original MR as well
 https://github.com/openshift-online/ocm-cli/pull/619#pullrequestreview-2144182593

Can we please sort this out now rather than later you are anyway touching all the same files 
, may be you can do it in a separate commit . 
I am not sure why we decided to let this go during the initial checkin, now seems like the time to get this right.

JakobGray:
The original MR addressed it by removing any panics and surfacing all errors to the top level function. This latest commit takes it a step further by removing all `log.Fatal`'s from the top level function as well.

PR #644, opened 2024-07-19
Add wif config cluster create option

JakobGray:
Issue: https://issues.redhat.com/browse/OCM-8008

- Adds --wif-config flag for cluster create
- Adds interactive prompt following CCS for authentication method

**Interactive view (select auth type):**
```
? CCS: Yes
? Method of authenticating GCP cluster
? Authentication method:  [Use arrows to move, type to filter]
> Workload Identity Federation (WIF)
  Service account
```

**Interactive view (select WIF config to use):**
```
? CCS: Yes
? Authentication method: Workload Identity Federation (WIF)
? Specifies the GCP Workload Identity Federation config used to authenticate.
? WIF Configuration:  [Use arrows to move, type to filter]
> jdg01 (2crdj6l18l0uqfoqq4873a4alp0t9v4r)
```

**Create cluster help view:**
```
Usage:
  ocm create cluster [flags] NAME

Flags:
      --wif-config string                                     Specifies the GCP Workload Identity Federation config used to authenticate.
```

### DIFF
--- a/pkg/provider/wif_configs.go
+++ b/pkg/provider/wif_configs.go
@@ -1,0 +1,56 @@
+package provider
+
+import (
+	"fmt"
+
+	"github.com/openshift-online/ocm-cli/pkg/arguments"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+)
+
+func getWifConfigs(client *cmv1.Client) (wifConfigs []*cmv1.WifConfig, err error) {
+	collection := client.GCP().WifConfigs()
+	page := 1
+	size := 100
+	for {
+		var response *cmv1.WifConfigsListResponse
+		response, err = collection.List().
+			Page(page).
+			Size(size).
+			Send()
+		if err != nil {
+			return
+		}
+		wifConfigs = append(wifConfigs, response.Items().Slice()...)
+		if response.Size() < size {
+			break
+		}
+		page++
+	}
+
+	if len(wifConfigs) == 0 {
+		return nil, fmt.Errorf("no WIF configurations available")
+	}
+	return
+}
+
+func GetWifConfigs(client *cmv1.Client) (wifConfigs []*cmv1.WifConfig, err error) {
+	return getWifConfigs(client)
+}
+
+// GetWifConfigNameOptions returns the wif config options for the cluster
+// with display name as the value and id as the description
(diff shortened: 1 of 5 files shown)


JakobGray:
@ckandag @renan-campos

pkg/provider/wif_configs.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	if len(wifConfigs) == 0 {
+		return nil, fmt.Errorf("no wif configs found")
+	}
+	return
+}
+
+func GetWifConfigs(client *cmv1.Client) (wifConfigs []*cmv1.WifConfig, err error) {
+	return getWifConfigs(client)
```
renan-campos, line 37: Out of curiosity: Why did you make place all of the functionality within a private helper?

pkg/provider/wif_configs.go
```diff
@@ -0,0 +1,56 @@
...
+			Size(size).
+			Send()
+		if err != nil {
+			return
+		}
+		wifConfigs = append(wifConfigs, response.Items().Slice()...)
+		if response.Size() < size {
+			break
+		}
+		page++
```
renan-campos, line 27: This is a silly edge case, but if the total happens to be equal to the size, then a call to page 2 will needlessly be made.  I think that the make the correct number of list calls would be to check if the accumulated number of resources seen so far is less then the total listed by the call. I think the way that is being done here is uniform with the rest of the codebase, so there's no need to fix it here.

renan-campos (approved):
LGTM. Going to do a quick run locally and then I'll merge it

pkg/provider/wif_configs.go
JakobGray: monkey see monkey do

pkg/provider/wif_configs.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	if len(wifConfigs) == 0 {
+		return nil, fmt.Errorf("no wif configs found")
```
ckandag, line 31: Is this the main error that user see when a wif cluster is creation is triggered but there are no wifconfigs ?

JakobGray: Yes

JakobGray: added syllables and changed to "available" instead of "found" since we intend to only display unused wif configs

PR #660, opened 2024-08-08
OCM-10207 | Filter wif configs in interactive mode

JakobGray:
Related issue: https://issues.redhat.com/browse/OCM-10207

Only display options for wif configs that are not used by any existing clusters.

Also accept wif config ID or name when assigning value to `--wif-config` flag

### DIFF
--- a/pkg/provider/wif_configs.go
+++ b/pkg/provider/wif_configs.go
@@ -7,7 +7,9 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 )
 
-func getWifConfigs(client *cmv1.Client) (wifConfigs []*cmv1.WifConfig, err error) {
+const unusedWifsQuery = "cluster.id is null"
+
+func getWifConfigs(client *cmv1.Client, filter string) (wifConfigs []*cmv1.WifConfig, err error) {
 	collection := client.GCP().WifConfigs()
 	page := 1
 	size := 100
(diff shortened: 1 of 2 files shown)


pkg/provider/wif_configs.go
```diff
@@ -7,7 +7,9 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 )
 
-func getWifConfigs(client *cmv1.Client) (wifConfigs []*cmv1.WifConfig, err error) {
+var unusedWifsQuery = "cluster.id is null"
```
renan-campos, line 10: QQ: why a var instead of a const?

pkg/provider/wif_configs.go
```diff
@@ -34,13 +37,38 @@ func getWifConfigs(client *cmv1.Client) (wifConfigs []*cmv1.WifConfig, err error
...
+		return nil, fmt.Errorf("there are %d WIF configurations found with identifier or name '%s'", len(wifs), key)
+	}
+	return wifs[0], nil
 }
 
 // GetWifConfigNameOptions returns the wif config options for the cluster
 // with display name as the value and id as the description
 func GetWifConfigNameOptions(client *cmv1.Client) (options []arguments.Option, err error) {
-	wifConfigs, err := getWifConfigs(client)
+	wifConfigs, err := getWifConfigs(client, unusedWifsQuery)
```
renan-campos, line 71: Minor, but I was wondering if it would be better for readability to call GetUnusedWifConfigs here?

renan-campos (approved):

pkg/provider/wif_configs.go
JakobGray: Changed to const

pkg/provider/wif_configs.go
JakobGray: Changed to GetUnusedWifConfigs

ckandag (approved):

renan-campos (approved):

PR #662, opened 2024-08-27
Add role prefix flag on create wif-config

JakobGray:
Related issue: https://issues.redhat.com/browse/OCM-10458

### DIFF
--- a/cmd/ocm/gcp/create-wif-config.go
+++ b/cmd/ocm/gcp/create-wif-config.go
@@ -27,10 +27,11 @@ import (
 var (
 	// CreateWifConfigOpts captures the options that affect creation of the workload identity configuration
 	CreateWifConfigOpts = options{
-		DryRun:    false,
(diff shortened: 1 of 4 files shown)


cmd/ocm/gcp/create-wif-config.go
```diff
@@ -43,6 +44,7 @@ const (
 func NewCreateWorkloadIdentityConfiguration() *cobra.Command {
 	createWifConfigCmd := &cobra.Command{
 		Use:     "wif-config",
+		Aliases: []string{"wif-configs"},
```
renan-campos, line 47: `ocm gcp create wif-configs`?

I don't really see a need for this alias...

renan-campos (changes requested):
Quick question on the alias, otherwise lgtm

renan-campos (approved):
/lgtm

PR #664, opened 2024-09-13
Revert "Grant access to support group during WifConfig creation"

renan-campos:
The change introduced race-condition related errors. Role-binding is attempted before GCP has had a chance to fully register the service account created. This results in BadRequest errors and the need to create a new WifConfig call.


cmd/ocm/gcp/create-wif-config.go
```diff
@@ -123,40 +93,58 @@ func createWorkloadIdentityConfigurationCmd(cmd *cobra.Command, argv []string) e
 		return nil
 	}
 
-	gcpClientWifConfigShim := NewGcpClientWifConfigShim(GcpClientWifConfigShimSpec{
-		GcpClient: gcpClient,
-		WifConfig: wifConfig,
-	})
-
-	if err := gcpClientWifConfigShim.GrantSupportAccess(ctx, log); err != nil {
```
JakobGray, line 131: Removing this line would be enough to resolve the issue until it can be properly addressed

PR #665, opened 2024-09-13
fix: tolerate required delay between service account creation and con…

renan-campos:
…figuration

It was discovered through testing that service accounts created on GCP need a duration of time between creation and being referenced, otherwise a BadRequest error occurs. A delayed retry logic is introduced to ensure the service account is available before making additional configuration calls.


JakobGray (approved):

JakobGray:
Addressed in https://github.com/openshift-online/ocm-cli/pull/666

PR #666, opened 2024-09-16
Wif creation improvements, including logic to grant support access as part of wif creation.

renan-campos:
These changes were pushed last week as part of https://github.com/openshift-online/ocm-cli/pull/663, but needed to be reverted due to a timing-related issue on the GCP side. 

The issue was investigated and determined to be as follows. The service account is created with an iam api call, but the binding of roles to the service account is made with a cloudresourcemanager api call. It was found that there is a window of time in which the service account created is not yet visible to cloudresourcemanager, resulting in sporadic BadRequest errors. This PR reintroduces the functionality with an additional mechanism to make wif creation robust to these out-of-sync issues.

Additionally there is an improvement to the logic used to determine if a custom role should be updated. I was finding that the osd_deployer_v4.17 role was getting updated every time I called wif create. This was caused by the list of permissions not being in alphabetical order. The new logic will only update the role if the existing and proposed permissions do not have the same elements, regardless of order.

To test this, the following bash command was run. After the changes, the BadRequest errors previously received were not experienced.

```
for id in $(awk 'BEGIN{for (i=1;i<10;i++) print "rc-"i}') ; do ./ocm gcp create wif-config --name $id --project $project_id; done
```


cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -0,0 +1,542 @@
...
+	// account creation call is made that the service account is not available
+	// in adjacent API calls. The call is therefore wrapped in retry logic to
+	// be robust to these types of synchronization issues.
+	return utils.DelayedRetry(func() error {
+		return c.bindRolesToPrincipal(
+			ctx,
+			fmt.Sprintf("serviceAccount:%s@%s.iam.gserviceaccount.com", serviceAccountId, c.wifConfig.Gcp().ProjectId()),
+			roles,
+		)
+	}, 10, 500*time.Millisecond)
```
ckandag, line 302: would there be benefit to making these consts ?

 how did we determine that is enough? 5 secs doesnt seem like much for syncing issues in GCP

JakobGray (approved):
lgtm. In the future I think we could improve the helper to only retry based on the type of GCP error we get

cmd/ocm/gcp/gcp-client-shim.go
renan-campos: I will move these to constants. This value was found empirically: I found it takes about 2 seconds for each service account to be configured.

PR #667, opened 2024-09-18
OCM-10615 | Implement 'gcp wif-config update' command

JakobGray:
1. Implement 'gcp wif-config update' command to re-apply GCP resources
2. Filter version on 'create cluster' when using WIF
3. Accept WIF displayName in WIF commands

Related issue: [OCM-10615](https://issues.redhat.com/browse/OCM-10615)


cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -545,3 +545,12 @@ func (c *shim) createMemberRoleBindingForProject(
 		Role:    roleName,
 	})
 }
+
+// Checks for WIF config name or id in input
+func wifKeyArgCheck(args []string) error {
```
renan-campos, line 550: This helper does not belong in this file. I would put it in a helpers.go file.

renan-campos, line 550: This helper does not belong in this file as it does not relate to the shim. I would recommend placing it in a 'helpers.go' file.

cmd/ocm/gcp/update-wif-config.go
```diff
@@ -1,39 +1,106 @@
...
+		return err
+	}
 	return nil
 }
 
 func updateWorkloadIdentityConfigurationCmd(cmd *cobra.Command, argv []string) error {
-	// No implementation yet
+	ctx := context.Background()
+	log := log.Default()
+	key := argv[0]
```
renan-campos, line 40: I have readability concerns with the way we are deriving the key throughout our wif commands. 

We know that the first element of argv exists and will be populated because we run wifKeyArgCheck on argv within the preRunE function that is run prior to RunE; we are safe to access argv[0]. However knowing this requires the reader to be aware of this execution flow, which is defined by the cobra package.

I'm thinking it may be better for the readibility of the code to do away with the preRunE call, and do the flag validation at the start of RunE.  We could then have functions that both validate that the data is there and return it:
```
key, err := wifKeyFromArgs(argv)
```

cmd/ocm/gcp/generate-wif-script.go
renan-campos, line 1: Should we have a manual mode for wif updates?

cmd/ocm/gcp/update-wif-config.go
```diff
@@ -1,39 +1,106 @@
...
+
+	if err = gcpClientWifConfigShim.CreateServiceAccounts(ctx, log); err != nil {
+		return fmt.Errorf("Failed to update IAM service accounts: %s", err)
+	}
+
 	return nil
 }
+
+// findWifConfig finds the WIF configuration by ID or name
+func findWifConfig(client *cmv1.Client, key string) (*cmv1.WifConfig, error) {
```
renan-campos, line 86: This function could also go in helpers.go, as it is used by multiple commands.

renan-campos (changes requested):
Requested changes pertain to code organization. The new function implementations and overall enhancements look good to me.

cmd/ocm/gcp/gcp-client-shim.go
JakobGray: Moved things to a helpers.go

cmd/ocm/gcp/update-wif-config.go
JakobGray: I moved to a helper function and took it out of the preRun validation

cmd/ocm/gcp/generate-wif-script.go
JakobGray: I added support to generate the script. It adds a new section to update existing IAM roles

renan-campos:
Couple things I found while testing this out:
1. The create.sh and apply.sh scripts have an error in them: looks like a space is missing between the project name and permissions.
2. The logging of update needs some attention.

- The sre role is getting updated every time it is called. 
- Granting of support is logged as occurring every time as well.
- Workload identity federation pool and provider are logged as "already exists" but the other resources are not.

renan-campos (dismissed):
See above comment.

renan-campos:
```
[rcampos@rcampos-thinkpadt14sgen2i ocm-cli]$ ./ocm gcp update wif-config rc-test
2024/09/25 13:27:45 Role "sre_managed_support" updated
2024/09/25 13:27:46 support access granted to sd-sre-platform-gcp-access@redhat.com
2024/09/25 13:27:46 Workload identity pool 2e0vgthhfsdl1kbrbokvakum7r8tqm24 exists
2024/09/25 13:27:46 Workload identity provider oidc exists
```

JakobGray:
I've modified the logging so it won't log anything if no updates are done. I changed the createOrUpdateRoles so it adds missing permissions rather than replacing them. And I fixed the gcloud command and tested that it can run properly.

I tested with 3 changes:
1. Deleted SA `osd-worker-g064@sda-ccs-3.iam.gserviceaccount.com`
2. Added a permission to role `osd_deployer_v4.17`
3. Removed a permission on role `openshift_gcp_pd_csi_driver_operator_v4.17`

Running results in these logs:
```
go run ./cmd/ocm gcp update wif-config kms
2024/09/26 12:41:50 IAM service account osd-worker-g064 created
2024/09/26 12:41:54 Role "openshift_gcp_pd_csi_driver_operator_v4.17" updated
```

The result is
1. `osd-worker-g064@sda-ccs-3.iam.gserviceaccount.com` is recreated
2. The added permission in `osd_deployer_v4.17` is left
3. The removed permission in `openshift_gcp_pd_csi_driver_operator_v4.17` is added back

ckandag:
I scanned to make sure all of the PR review comments were addressed.
I locally verified there is no regression in the green paths i.e creation of a wif , creation of a basic gcp cluster 
also verified version filtering for wif is working as expected.

I have not tested the scripts though.  I am merging this MR but @JakobGray  pls make sure to verify all the dry run scripts and post a comment here and address any issues post-merge

ckandag (approved):

PR #670, opened 2024-10-01
Dry-run wif config delete before tearing down cloud resources

JakobGray:
Issue: [OCM-11561](https://issues.redhat.com/browse/OCM-11561)

Dry-run the delete to verify the wif-config can be deleted so the WIF resources aren't destroyed while in use.

### DIFF
--- a/cmd/ocm/gcp/delete-wif-config.go
+++ b/cmd/ocm/gcp/delete-wif-config.go
@@ -70,6 +70,13 @@ func deleteWorkloadIdentityConfigurationCmd(cmd *cobra.Command, argv []string) e
 		return errors.Wrapf(err, "failed to get wif-config")
 	}
 
+	// Check if wif-config can be deleted with dry-run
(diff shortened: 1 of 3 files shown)


cmd/ocm/gcp/delete-wif-config.go
```diff
@@ -80,6 +80,13 @@ func deleteWorkloadIdentityConfigurationCmd(cmd *cobra.Command, argv []string) e
 		return nil
 	}
 
+	// Check if wif-config can be deleted with dry-run
+	_, err = connection.ClustersMgmt().V1().GCP().WifConfigs().
+		WifConfig(wifConfig.ID()).Delete().DryRun(true).Send()
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete wif config %q", wifConfig.ID())
+	}
+
```
renan-campos, line 89: This check should occur prior to the servicing of cli's dry-run. We don't want the user to generate the cleanup script if the cluster is in use.

renan-campos (approved):

PR #678, opened 2024-10-17
OCM-11703 | Update provider flag check 

JakobGray:
Issue: https://issues.redhat.com/browse/OCM-11703

Require provider, do not default to AWS, and check for provider-specific flags

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -615,9 +615,6 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		return err
 	}
 
-	// Only offer the 2 providers known to support OSD now;
-	// but don't validate if set, to not block `ocm` CLI from creating clusters on future providers.
-	providers, _ := osdProviderOptions(connection)
 	// If marketplace-gcp subscription type is used, provider can only be GCP
 	gcpBillingModel, _ := billing.GetBillingModel(connection, billing.MarketplaceGcpSubscriptionType)
 	gcpSubscriptionTypeTemplate := subscriptionTypeOption(gcpBillingModel.ID(), gcpBillingModel.Description())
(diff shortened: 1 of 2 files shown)


cmd/ocm/create/cluster/cmd.go
```diff
@@ -912,6 +918,22 @@ func promptName(argv []string) error {
...
+func promptProvider(fs *pflag.FlagSet, connection *sdk.Connection) error {
+	// Only offer the 2 providers known to support OSD now;
+	// but don't validate if set, to not block `ocm` CLI from creating clusters on future providers.
+	providers, _ := osdProviderOptions(connection)
+
+	err := arguments.PromptOneOf(fs, "provider", providers)
+	if err != nil {
+		return err
+	}
+	if args.provider == "" {
```
ckandag, line 931: Given that provider is inferred for the case where subscription is gcpMarketPlace,
 i.e provider  is only required if subscription is not gcpMarketPlace,  
We cannot mark the provider flag as required (the Cmd.MarkFlagRequired)  and instead we are doing a  manual check  if !gcpMarketPlace

pkg/arguments/arguments.go
```diff
@@ -157,6 +157,52 @@ func CheckIgnoredCCSFlags(ccs cluster.CCS) error {
...
+	}
+	if provider != cluster.ProviderAWS {
+		for _, flag := range awsExclusiveFlags {
+			if fs.Changed(flag) {
+				bad = append(bad, flag)
+			}
+		}
+	}
+	if len(bad) == 1 {
+		return fmt.Errorf("%s flag is meaningless using chosen provider", bad[0])
```
ckandag, line 198: shouldnt it be.  `flag is meaningless "when" using chosen provider`

ckandag:
please review err msg , looks good otherwise

ckandag (approved):

pkg/arguments/arguments.go
JakobGray: Updated verbiage

PR #681, opened 2024-10-21
OCM-11842 : Improve gcp cluster creation with psc -UX

ckandag:
https://issues.redhat.com/browse/OCM-11842

1. Make the interactive workflow for ClusterPrivacy/ BYO- VPC  similar to UI
2. For WIF clusters, Ensure  PSC is set when Cluster is Private

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -52,6 +52,12 @@ const (
 	gcpTermsAgreementNonInteractiveError = "Review and accept Google Terms and Agreements on " +
 		gcpTermsAgreementsHyperlink + ". Set the flag --marketplace-gcp-terms to true " +
 		"once agreed in order to proceed further."
+
+	privateFlag            = "private"
+	vpcNameFlag            = "vpc-name"
+	controlPlaneSubnetFlag = "control-plane-subnet"
+	computePlaneSubnetFlag = "compute-subnet"
+	pscSubnetFlag          = "psc-subnet"
 )
 
 var args struct {
@@ -234,7 +240,7 @@ func init() {
 	fs.MarkHidden("expiration")
 	fs.BoolVar(
 		&args.private,
-		"private",
+		privateFlag,
 		false,
 		"Restrict master API endpoint and application routes to direct, private connectivity.",
 	)
@@ -379,11 +385,11 @@ func init() {
 
 	fs.StringVar(
 		&args.gcpPrivateSvcConnect.SvcAttachmentSubnet,
-		"psc-subnet",
+		pscSubnetFlag,
 		"",
 		"Specifies the ServiceAttachment Subnet for Private Service Connect in GCP",
 	)
-	arguments.SetQuestion(fs, "psc-subnet", "PrivatSericeConnect ServiceAttachment Subnet:")
+	arguments.SetQuestion(fs, pscSubnetFlag, "PrivateServiceConnect ServiceAttachment Subnet:")
 
 	fs.StringVar(
 		&args.gcpAuthentication.Type,
@@ -748,11 +754,21 @@ func preRun(cmd *cobra.Command, argv []string) error {
 		}
 	}
 
+	err = promptClusterPrivacy(fs)
+	if err != nil {
+		return err
+	}
+
 	if args.existingVPC.SubnetIDs != "" {
 		args.existingVPC.Enabled = true
 	}
 
 	err = promptExistingVPC(fs, connection)
+	if err != nil {
+		return err
+	}
+
+	err = promptPrivateServiceConnect(fs)
 	if err != nil {
 		return err
 	}
@@ -768,11 +784,6 @@ func preRun(cmd *cobra.Command, argv []string) error {
 	}
 
 	err = promptNetwork(fs)
-	if err != nil {
-		return err
-	}
-
-	err = promptPrivateServiceConnect(fs)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


cmd/ocm/create/cluster/cmd.go
```diff
@@ -1334,6 +1339,10 @@ func promptExistingVPC(fs *pflag.FlagSet, connection *sdk.Connection) error {
 	return err
 }
 
+func promptClusterPrivacy(fs *pflag.FlagSet, connection *sdk.Connection) error {
+	return arguments.PromptBool(fs, "private")
```
renan-campos, line 1343: Considering that this string value must match the string value of the flag definition, it would be an improvement to maintainability if it were a const instead of a hard-coded value.

cmd/ocm/create/cluster/cmd.go
```diff
@@ -1500,7 +1506,11 @@ func promptPrivateServiceConnect(fs *pflag.FlagSet) error {
 		!args.existingVPC.Enabled || !args.private {
 		return nil
 	}
-	isPSC := (args.gcpPrivateSvcConnect.SvcAttachmentSubnet != "")
+
+	//if Wif cluster and private is enabled then has to be PSC
```
renan-campos, line 1510: Why is that? I wasn't aware of this.

cmd/ocm/create/cluster/cmd.go
```diff
@@ -1519,7 +1529,10 @@ func promptPrivateServiceConnect(fs *pflag.FlagSet) error {
 			return err
 		}
 	}
-
+	if isWif && args.gcpPrivateSvcConnect.SvcAttachmentSubnet == "" {
+		return fmt.Errorf(
+			"flag 'psc-subnet' is required when cluster is 'private' and GCP authentication type is %s", c.AuthenticationWif)
```
renan-campos, line 1534: 'psc-subnet' and 'private' should be consts instead of hard-coded values for maintainability purposes. By using consts, the value need only be changed in one location to impact all parts of the code that use it. Hard-coded values that must be identifcal in multiple sections of the code rely on the modified to ensure they are thorough when changing the value, which opens a potential for defects as the number of locations increase.

renan-campos (changes requested):

JakobGray (changes requested):
I did a dry-run flow with this branch and got the following
```
go run ./cmd/ocm create cluster -i --dry-run=true 
? cluster name jjjj
? Subscription type: standard (Annual: Fixed capacity subscription from Red Hat)
? Cloud provider: gcp
? CCS: Yes
? Authentication method: Workload Identity Federation (WIF)
? WIF Configuration: jg1021 (2ehvi64du4ahq77a63vf3gj70bm63bpa)
? Multiple AZ: No
? Secure boot support for Shielded VMs: No
? Region: us-east1
? OpenShift version: 4.17.0
? Compute machine type: n2-standard-4
? Enable autoscaling: No
? Compute nodes: 2
? Private cluster (optional): Yes
? Install into an existing VPC (optional): No
? Machine CIDR: 10.0.0.0/16
? Service CIDR: 172.30.0.0/16
? Pod CIDR: 10.128.0.0/14
? Host prefix: 23
? Domain Prefix: 
dry run: Would be successful.
```

Based on point 2 I should not be able to select `No` for the prompt `Install into an existing VPC (optional)`

Also there are typos in the PSC prompt
```
? PrivatSericeConnect ServiceAttachment Subnet: [? for help] 
```

cmd/ocm/create/cluster/cmd.go
```diff
@@ -52,6 +52,12 @@ const (
 	gcpTermsAgreementNonInteractiveError = "Review and accept Google Terms and Agreements on " +
 		gcpTermsAgreementsHyperlink + ". Set the flag --marketplace-gcp-terms to true " +
 		"once agreed in order to proceed further."
+
+	privateFlag            = "private"
+	vpcNameFlag            = "vpc-name"
+	controlPlaneSubnetFlag = "compute-subnet"
+	computePlaneSubnetFlag = "control-plane-subnet"
```
renan-campos, line 59: These two const names are mixed up for their values and not used in defining the flag.

cmd/ocm/create/cluster/cmd.go
```diff
@@ -1294,15 +1332,11 @@ func promptExistingGCPVPC(fs *pflag.FlagSet, connection *sdk.Connection) error {
...
 				args.existingVPC.ComputeSubnet)
 		}
 	}
 
-	if wasGCPNetworkReceived() {
-		args.existingVPC.Enabled = true
-	}
-
 	fs.Set("use-existing-vpc", "true")
 	flag := fs.Lookup("vpc-name")
```
renan-campos, line 1341:
```suggestion
	flag := fs.Lookup(vpcNameFlag)
```

cmd/ocm/create/cluster/cmd.go
renan-campos: If we are adding consts for these, we should replace the hard-coded values.

renan-campos (changes requested):

cmd/ocm/create/cluster/cmd.go
ckandag: I just replaced in the code I  was working on but yes I replace all

ckandag:
> My comment is still mostly the same
> 
> ```
> ocm create cluster -i --dry-run=true --ccs  --provider=gcp --region=us-east1 --compute-machine-type=n2-standard-4 --psc-subnet=jdg-21-tf2d8-worker-subnet jjjj
> ? Subscription type: standard (Annual: Fixed capacity subscription from Red Hat)
> ? Authentication method: Service account
> ? Service account file: /Users/jdgray/gitlab/temps/sda-ccs-1/jdg-ccs-1-key.json
> ? Multiple AZ: No
> ? Secure boot support for Shielded VMs: No
> ? OpenShift version: 4.17.0
> ? Enable autoscaling: No
> ? Compute nodes: 2
> ? Private cluster (optional): Yes
> ? Install into an existing VPC (optional): [? for help] (y/N) 
> ```
> 
> In the interactive flow where I've already indicated that I'm using PSC it asks me if I want a private cluster and if I want to install into an existing VPC. Both of these prompts have a "right" and "wrong" answer in this case, and the default option is the incorrect one.
> 
> If managing the flow is too complex then it might not be worth pursuing, because at least if the user chooses wrong they receive quick feedback on their mistake.

The interactive flow I think has value and does work when user tries it all interatcively. You are trying an edge case with a mix of both non interactive and interactive flags which is breaking it.
I do not think we do remove the entire interactive flow jsut so we can satisfy an edge case

JakobGray (approved):

cmd/ocm/create/cluster/cmd.go
renan-campos, line 59:
```suggestion
	controlPlaneSubnetFlag = "control-plane-subnet"
	computePlaneSubnetFlag = "compute-subnet"
```

ckandag: sorry, I missed amending my commit before pushing the changes

renan-campos (approved):
/lgtm

Issue #128, opened 2020-08-26
Make all cluster ID formats usable in all ocm commands


mmazur:
There are three cluster id formats in use:
* ocm id: 1egguj87evf1u1adjj8nj8p25n06i0fd
* external id: e5666af8-d3f5-4524-b8db-44607c243b35
* name: qwerty.asdf.p1

The expectation would be that all commands that operate on clusters would accept all of the above without complaining. That's unfortunately not the case.

PR #696, opened 2024-12-11
OCM-12971 | 'ocm gcp update wif-config' remediates all wif-config misconfigurations

renan-campos:
**Overview**
The following new behavior was added to the functionality of the command used to update resources associated with wif-configs:
- re-enable disabled service accounts
- re-enable disabled custom roles
- re-enable disabled workload identity pool
- overwrite identity provider misconfigurations

**Testing**
The resources of a test wif-config resource was manually modified into the states described above. Running `ocm gcp verify wif-config` confirmed that the error related to the misconfiguration was detected. After running `ocm gcp update wif-config`, verification succeeds. Manual inspection of GCP resources confirms that their configuration was corrected through calling the update command.

### DIFF
--- a/cmd/ocm/gcp/gcp-client-shim.go
+++ b/cmd/ocm/gcp/gcp-client-shim.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"reflect"
 	"sort"
 	"strings"
 	"time"
@@ -54,22 +55,16 @@ func (c *shim) CreateWorkloadIdentityPool(
 	ctx context.Context,
 	log *log.Logger,
 ) error {
+	description := fmt.Sprintf(wifDescription, c.wifConfig.DisplayName())
 	poolId := c.wifConfig.Gcp().WorkloadIdentityPool().PoolId()
 	project := c.wifConfig.Gcp().ProjectId()
 
 	parentResourceForPool := fmt.Sprintf("projects/%s/locations/global", project)
 	poolResource := fmt.Sprintf("%s/workloadIdentityPools/%s", parentResourceForPool, poolId)
+
 	resp, err := c.gcpClient.GetWorkloadIdentityPool(ctx, poolResource)
-	if resp != nil && resp.State == "DELETED" {
-		log.Printf("Workload identity pool %s was deleted", poolId)
-		_, err := c.gcpClient.UndeleteWorkloadIdentityPool(
-			ctx, poolResource, &iamv1.UndeleteWorkloadIdentityPoolRequest{},
-		)
-		if err != nil {
-			return errors.Wrapf(err, "failed to undelete workload identity pool %s", poolId)
-		}
-	} else if err != nil {
-		description := fmt.Sprintf(wifDescription, c.wifConfig.DisplayName())
+
+	if err != nil {
 		if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 404 &&
 			strings.Contains(gerr.Message, "Requested entity was not found") {
 			pool := &iamv1.WorkloadIdentityPool{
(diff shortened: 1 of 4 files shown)


cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -90,42 +91,55 @@ func (c *shim) CreateWorkloadIdentityPool(
 		}
 	}
 
+	// Enable the pool if it exists but is disabled.
+	if resp.Disabled {
```
ckandag, line 95: resp != nil && resp.Disabled

cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -90,42 +91,55 @@ func (c *shim) CreateWorkloadIdentityPool(
...
+		if err := c.gcpClient.EnableWorkloadIdentityPool(
+			ctx,
+			resp.Name,
+		); err != nil {
+			return err
```
ckandag, line 100: should this not be wrapped similar to line #90

cmd/ocm/gcp/gcp-client-shim.go
ckandag: this if condition should either be incorporate into previous if\else    or  make those previous if\else return after those conditions are processed sucessfully. This just seemed mixed up

pkg/gcp/client.go
```diff
@@ -178,6 +182,34 @@ func (c *gcpClient) DeleteWorkloadIdentityPool(ctx context.Context, resource str
...
+	ctx context.Context,
+	poolId string,
+) error {
+	_, err := c.oldIamClient.Projects.Locations.WorkloadIdentityPools.Patch(
+		poolId,
+		&iamv1.WorkloadIdentityPool{
+			Disabled: false,
+		},
+	).UpdateMask("disabled").Do()
+	return err
```
ckandag, line 210: does this err not have to be handled same as  other error msgs  line @ line#195

pkg/gcp/client.go
```diff
@@ -233,3 +265,14 @@ func (c *gcpClient) UndeleteWorkloadIdentityPool(ctx context.Context, resource s
...
+
+func (c *gcpClient) UpdateWorkloadIdentityPoolOidcIdentityProvider(
+	ctx context.Context,
+	provider *iamv1.WorkloadIdentityPoolProvider,
+) error {
+	_, err := c.oldIamClient.Projects.Locations.WorkloadIdentityPools.Providers.Patch(
+		provider.Name,
+		provider,
+	).UpdateMask("attributeMapping,description,displayName,disabled,state,oidc").Do()
+	return err
```
ckandag, line 277: does this err not have to be handled same as other error like at @ line#195

cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -139,6 +153,38 @@ func (c *shim) CreateWorkloadIdentityProvider(
 		}
 	}
 
+	var needsUpdate bool
```
ckandag, line 156: below is needed only if the prev if check at #128 fails.

So I think you need a `return nil ` after line#149 , so it doesnt get into the below

cmd/ocm/gcp/gcp-client-shim.go
renan-campos: I have simplified the conditional logic of this method based on your comment.

cmd/ocm/gcp/gcp-client-shim.go
renan-campos: ack.

pkg/gcp/client.go
renan-campos: ack.

pkg/gcp/client.go
renan-campos: ack.

ckandag (approved):

PR #698, opened 2024-12-18
OCM-12391 | feat: Add 'version' flag to wif-config create and update commands

JakobGray:
Version flag expects a string representing an Openshift minor release in the form 'X.Y'. On create this sets the wif template used by the wif config. On upgrade, the version is added to the list of templates.

Given a version like `4.17`, the CLI will parse that it is in `X.Y` form and convert it to the expected wif template id form `v4.17`. Otherwise it will pass the input directly for the backend to validate.

Example use:
```
➜  ocm-cli git:(OCM-12391) ✗ go run ./cmd/ocm gcp create wif-config --name=jdg3 --project=sda-ccs-1 --version=4.17 --mode=manual
```

The flag can also be set during interactive create mode:
```
➜  ocm-cli git:(OCM-12391) ✗ go run ./cmd/ocm gcp create wif-config --mode=manual -i
? wif-config name: name
? Gcp Project ID: sda-ccs-1
? The OCP version to configure the wif-config for. Will default to the latest supported version if left unset.
? Openshift version: 
```

**Manual Mode**
Script generated by manually running an update with two versions: [apply.txt](https://github.com/user-attachments/files/18187687/apply.txt)
This doubles the number of roles to create and update

**Tests run**
- Returning 400 when giving an invalid version input
- Update adds new roles for version, and old roles from existing version if missing
- Update generates script covering both versions assigned
- Verify command passes following an update

**Issues:** 
[OCM-12391](https://issues.redhat.com/browse/OCM-12391)
[OCM-12396](https://issues.redhat.com/browse/OCM-12396)

### DIFF
--- a/cmd/ocm/gcp/create-wif-config.go
+++ b/cmd/ocm/gcp/create-wif-config.go
@@ -19,11 +19,12 @@ import (
 var (
 	// CreateWifConfigOpts captures the options that affect creation of the workload identity configuration
 	CreateWifConfigOpts = options{
-		Mode:       ModeAuto,
-		Name:       "",
-		Project:    "",
-		RolePrefix: "",
-		TargetDir:  "",
+		Mode:             ModeAuto,
+		Name:             "",
+		Project:          "",
+		RolePrefix:       "",
+		TargetDir:        "",
+		OpenshiftVersion: "",
 	}
 )
 
@@ -75,6 +76,12 @@ wif-config resource within OCM to represent those resources.`,
 		"",
 		targetDirFlagDescription,
 	)
+	createWifConfigCmd.PersistentFlags().StringVar(
+		&CreateWifConfigOpts.OpenshiftVersion,
+		"version",
+		"",
+		versionFlagDescription,
+	)
 
 	return createWifConfigCmd
 }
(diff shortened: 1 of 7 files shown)


cmd/ocm/gcp/create-wif-config.go
```diff
@@ -137,6 +147,23 @@ func promptProjectId() error {
 	return nil
 }
 
+func promptVersion() error {
+	const projectIdHelp = "The OCP version to configure the wif-config for."
```
renan-campos, line 151:
```suggestion
	const versionHelp = "The OCP version to configure the wif-config for."
```
I believe this may have originally been a name that was copied from promptProjectId. I suggest renaming it to improve readibility.

renan-campos: We should specify in the help message that the user may leave the field empty to use the latest wif version.

renan-campos (commented):
There is a constant name that should be fixed.  We must communicate to the user that leaving the version option empty during wif-config creation will result in a wif-config that uses the latest version.

renan-campos (approved):
Tested in local environment. LGTM

PR #700, opened 2025-01-04
Improved UX for updating wif-config.

renan-campos:
The included changes provide the following properties to the update command: 
* Operations that modify GCP cloud resources are always accompanied by a log message displayed to the user.
* Changes to GCP cloud resources are made only when necessary.

Additional logic was needed to check the configuration of jwks and service account access. Prior to this, these resources were being updated every time the command was called.

### DIFF
--- a/cmd/ocm/gcp/gcp-client-shim.go
+++ b/cmd/ocm/gcp/gcp-client-shim.go
@@ -27,6 +27,13 @@ const (
 	retryDelayMs = 500
 )
 
+const (
+	impersonatorRole         = "roles/iam.serviceAccountTokenCreator"
+	workloadIdentityUserRole = "roles/iam.workloadIdentityUser"
+)
+
+// All operations that modify cloud resources should be logged to the user.
+// For this reason, all methods of this interface take a logger as a parameter.
 type GcpClientWifConfigShim interface {
 	CreateServiceAccounts(ctx context.Context, log *log.Logger) error
 	CreateWorkloadIdentityPool(ctx context.Context, log *log.Logger) error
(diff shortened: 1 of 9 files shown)


JakobGray (approved):

cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -542,18 +545,19 @@ func (c *shim) undeleteRole(
...
 	projectName string,
-) error {
+) (bool, error) {
 	needPolicyUpdate := false
 
 	policy, err := c.gcpClient.GetProjectIamPolicy(ctx, projectName, &cloudresourcemanager.GetIamPolicyRequest{})
 
 	if err != nil {
-		return fmt.Errorf("error fetching policy for project: %v", err)
+		return false, fmt.Errorf("error fetching policy for project: %v", err)
```
tzvatot, line 560: Not related specifically to this MR: why not using errors.Wrap (as done in other places in this file)?

pkg/gcp/error_handlers.go
```diff
@@ -8,23 +8,14 @@ import (
 	"google.golang.org/grpc/codes"
 )
 
-func (c *gcpClient) handleAttachImpersonatorError(err error) error {
+func (c *gcpClient) handleApiError(err error) error {
 	pApiError, ok := err.(*apierror.APIError)
 	if !ok {
 		return fmt.Errorf("Unexpected error")
```
tzvatot, line 14: Not related to his MR: we should add some more context to this error - it's not helping as it is. I would assume it *should* never happen though.

pkg/utils/jwks.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Checks if two string parameters represent equal json web key sets. false is
+// returned if the two jwks do not have equivalent values, or if there is an
+// error processing the expected fields of either parameter.
+func JwksEqual(
+	jwksStrA string,
+	jwksStrB string,
+) bool {
+	var jwksA, jwksB struct {
+		Keys []struct {
```
tzvatot, line 16: What about `x5c` and `x5t`?

Is there a a third party comparison tool that we can use? trying not to re-invent the wheel

tzvatot (changes requested):

pkg/gcp/error_handlers.go
renan-campos: It should never happen. The call should return a google api error which the method is able to convert properly.

pkg/utils/jwks.go
renan-campos: updated to use https://pkg.go.dev/github.com/MicahParks/jwkset#JWKSMarshal

ckandag (approved):

tzvatot:
LGTM

PR #703, opened 2025-01-20
added 'availability-zone' argument to machine pool creation

renan-campos:
Users belonging to organizations in which this feature is enabled may specify an availability zone during machine pool creation for osd-gcp clusters.

The attached text file demonstrates some tests that were run to verify the behavior of this patch.
[cli-test.txt](https://github.com/user-attachments/files/18482711/cli-test.txt)


cmd/ocm/create/machinepool/cmd.go
```diff
@@ -80,84 +84,136 @@ func init() {
...
-	if len(argv) < 1 || argv[0] == "" {
+	fmt.Printf("Machine pool '%s' created on cluster '%s'\n", machinePoolId, args.ClusterKey)
+	return nil
+}
+
+func getCluster(
+	connection *sdk.Connection,
+	clusterKey string,
+) (ocm.Cluster, error) {
+	clusterData, err := c.GetCluster(connection, clusterKey)
```
ckandag, line 171: Looking like the check to verify clusterKey was removed and there is no other check done to see if clusterKey value was input at all by the user (here or in the calling function "run") before try to fetch the cluster from the backend.

There  should be a check to determine if that option is atleast set and if not, user should be prompted to provide a value rather than going to the backend,  You would still get an error from the backend but that call firstly isn't needed and the error msg prompted to the user wouldn't be correct either.

cmd/ocm/create/machinepool/cmd.go
```diff
@@ -175,72 +231,90 @@ func run(cmd *cobra.Command, argv []string) error {
...
 		Taints(taintBuilders...)
 
-	if len(args.additionalSecurityGroupIds) != 0 {
-		for i, sg := range args.additionalSecurityGroupIds {
-			args.additionalSecurityGroupIds[i] = strings.TrimSpace(sg)
+	if len(args.AdditionalSecurityGroupIds) != 0 {
+		for i, sg := range args.AdditionalSecurityGroupIds {
+			args.AdditionalSecurityGroupIds[i] = strings.TrimSpace(sg)
 		}
 		mpBuilder.AWS(
```
ckandag, line 282: This looks like is missing a provider check. I know this wasn't introduced in this PR but if its not something user should be setting for GCP or should be setting it only for AWS, it would be good to add that check as well.

renan-campos:
Outputs with latest updates:
```
[rcampos@rcampos-thinkpadt14sgen2i ocm-cli]$ ./ocm create machine-pool day2 --cluster ""   --availability-zone us-east1-b --instance-type "m
5.4xlarge" --replicas 1
Error: Failed to get cluster '': Cluster name, identifier or external identifier '' isn't valid: it must contain only letters, digits, dashe
s and underscores
```

Attempting to set security groups on a GCP cluster:
```
[rcampos@rcampos-thinkpadt14sgen2i ocm-cli]$ ./ocm list clusters
ID                                NAME                                                    API URL                                           
          OPENSHIFT_VERSION   PRODUCT ID  HCP      CLOUD_PROVIDER  REGION ID       STATE        
2ge6n9afq8st64j5lvjsvn71d0p2p44g  rc-test-gcp-multi-az                                    https://example.com/veryfakeapi                   
          NONE                osd         false    gcp             us-east1        ready        
2ge6ntsr2d7t8r7i5ht1fn9dn1s3q7cl  rc-test-gcp-single-az                                   https://example.com/veryfakeapi                   
          NONE                osd         false    gcp             us-east1        ready        
2ge7ac7mha6lnmduclheoj69v3vp1mmh  fake-rosa-cluster                                       NONE                                              
          NONE                osd         false    aws             us-east-1       ready        
[rcampos@rcampos-thinkpadt14sgen2i ocm-cli]$ ./ocm create machine-pool bad-mp --cluster 2ge6n9afq8st64j5lvjsvn71d0p2p44g --availability-zone
 us-east1-b --instance-type "n2-standard-48" --replicas 1 --additional-security-group-ids foobar
Error: 'additional-security-group-ids' may only be set for clusters using the 'aws' cloud provider.
```

PR #723, opened 2025-03-14
OCM-14358 | unbind service accounts when deleting wif-configs

renan-campos:
Deleted service accounts retain their project iam policy bindings. Without unbinding the service accounts that associated with wif-configs, users run the risk of hitting the GCP quota of 1,500 bindings and needing to manually delete these stale policy entries. Through the changes in this commit, service account bindings related to the wif-config being deleted will be unbound.

These changes were confirmed through a combination of unit tests and manual integration tests. The unit tests demonstrate that the logic to remove members from a binding works as expected. 

The correctness of the patch was tested as follows. The count of deleted service account bindings was first tallied. Then a wif-config resource was created. The wif-config resource was then deleted, and a new count of the number of deleted service account bindings was collected. Through this experiment we can see that the count of deleted service accounts does not change. Running the same test with the old cli results in the number of deleted service account bindings going up when the wif-config is deleted.

The attached text file has the log showing the test and the call used to count the  current number of service accounts.

[OCM-14358-test.txt](https://github.com/user-attachments/files/19238800/OCM-14358-test.txt)

### DIFF
--- a/cmd/ocm/gcp/gcp-client-shim.go
+++ b/cmd/ocm/gcp/gcp-client-shim.go
@@ -305,7 +305,7 @@ func (c *shim) createOrUpdateRoles(
 		roleID := role.RoleId()
 		roleTitle := role.RoleId()
 		permissions := role.Permissions()
-		existingRole, err := c.getRole(ctx, c.fmtRoleResourceId(role))
+		existingRole, err := c.getRole(ctx, c.formatRoleResourceId(role))
 		if err != nil {
 			if gerr, ok := err.(*apierror.APIError); ok && gerr.GRPCStatus().Code() == codes.NotFound {
 				_, err = c.createRole(
@@ -328,7 +328,7 @@ func (c *shim) createOrUpdateRoles(
 
 		// Undelete role if it was deleted
 		if existingRole.Deleted {
-			_, err = c.undeleteRole(ctx, c.fmtRoleResourceId(role))
+			_, err = c.undeleteRole(ctx, c.formatRoleResourceId(role))
 			if err != nil {
 				return errors.Wrap(err, fmt.Sprintf("Failed to undelete custom role '%s'", roleID))
 			}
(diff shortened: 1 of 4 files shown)


cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -445,6 +438,27 @@ func (c *shim) bindRolesToPrincipal(
...
+	roles []*cmv1.WifRole,
+) error {
+	modified, err := c.removePolicyBindingsFromProject(
+		ctx,
+		c.formatRoles(roles),
+		principal,
+		c.wifConfig.Gcp().ProjectId(),
+	)
+	if err != nil {
+		return errors.Errorf("Failed to bind roles to principal %s: %s", principal, err)
```
ckandag, line 454:
```suggestion
		return errors.Errorf("Failed to unbind roles to principal %s: %s", principal, err)
```

ckandag: Failed to unbind roles from principal

cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -607,13 +672,30 @@ func (c *shim) addPolicyBindingForProject(
...
-
 	// if we didn't find an existing binding entry, then make one
 	c.createMemberRoleBindingForProject(policy, roleName, memberName)
-
 	return true
 }
 
+// Applies the passed in function to the policy entries which have the desired
+// roles and members. The application function will return a boolean indicating
+// whether a modification to the binding was made. The method returns wether a
```
ckandag, line 682:
```suggestion
// whether a modification to the binding was made. The method returns whether a
```

cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -607,13 +672,30 @@ func (c *shim) addPolicyBindingForProject(
 			return c.addMemberToBindingForProject(memberName, policy.Bindings[i])
```
ckandag, line 672: this I guess , can we replaced with applyFunc

renan-campos: ty. updated.

ckandag (approved):

PR #860, opened 2025-08-27
support resource-scoped permissions wif-config

miguelhbrito:


cmd/ocm/gcp/scripting.go
```diff
@@ -263,8 +264,19 @@ func addRoleBindingsScript(wifConfig *cmv1.WifConfig) string {
 			} else {
 				roleResource = fmt.Sprintf("projects/%s/roles/%s", project, role.RoleId())
 			}
-			sb.WriteString(fmt.Sprintf("gcloud projects add-iam-policy-binding %s --member=%s --role=%s\n",
-				project, member, roleResource))
+			if role.ResourceBindings() != nil {
+				for _, resourceBinding := range role.ResourceBindings() {
+					if resourceBinding.Type() != serviceAccountType {
```
ckandag, line 269: we should consider atleast printing out an error if its an unexpected/unsupported binding type 
          switch resourceBinding.Type() {
          case "iam.serviceAccounts":
					saResourceId := gcp.FmtSaResourceId(resourceBinding.Name(), project)
					sb.WriteString(fmt.Sprintf("gcloud iam service-accounts add-iam-policy-binding %s --member=%s --role=%s\n",
						saResourceId, member, roleResource))
          default:
              log.Printf("Warning: unsupported resource type '%s' for resource '%s'",
                  resourceBinding.Type(), resourceBinding.Name())
              continue
          }

cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -678,11 +798,16 @@ func (c *shim) ensurePolicyBindingsForProject(
...
+		if definedRole.ResourceBindings() != nil {
+			continue
+		}
+		roleResourceId := c.formatRoleResourceId(definedRole)
+		roleBindingAdded := c.ensureRoleBindingInPolicy(policy, roleResourceId)
 		if roleBindingAdded {
 			needPolicyUpdate = true
 		}
-		memberAddedToRoleBinding := c.applyMemberToRoleInPolicy(policy, definedRole, member, c.addMemberToBindingForProject)
+		memberAddedToRoleBinding := c.applyMemberToRoleInPolicy(policy, roleResourceId, member, c.addMemberToBindingForProject)
```
ckandag, line 810: pls check and fix lint errors

cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -29,6 +29,7 @@ const (
 const (
 	impersonatorRole         = "roles/iam.serviceAccountTokenCreator"
 	workloadIdentityUserRole = "roles/iam.workloadIdentityUser"
+	serviceAccountType       = "iam.serviceAccounts"
```
ckandag, line 32:
```suggestion
	serviceAccountResourceType       = "iam.serviceAccounts"
```

cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -509,16 +520,17 @@ func (c *shim) bindRolesToPrincipal(
 ) error {
 	modified, err := c.ensurePolicyBindingsForProject(
 		ctx,
-		c.formatRoles(roles),
+		roles,
```
ckandag, line 523: why did you make this change?

cmd/ocm/gcp/gcp-client-shim.go
```diff
@@ -562,6 +597,91 @@ func (c *shim) grantAccessToServiceAccount(
 	return nil
 }
 
+func (c *shim) bindResourcesToServiceAccount(
+	ctx context.Context,
+	log *log.Logger,
+	serviceAccount *cmv1.WifServiceAccount,
+) error {
+	for _, role := range serviceAccount.Roles() {
```
ckandag, line 605: we have to consider the possibility that the role may not be pre-defined , so will need to be created before you can bind

cmd/ocm/gcp/gcp-client-shim.go
miguelhbrito: to make this if "definedRole.ResourceBindings()" in function "ensurePolicyBindingsForProject" 
if I let this as "c.formatRoles(roles)", it returns []string

cmd/ocm/gcp/gcp-client-shim.go
miguelhbrito: line 641, "c.formatRoleResourceId(role)", as parameter to func "c.bindRoleToResource"

rcampos2029 (commented):
The new method names are difficult to understand. I suggest the following naming scheme:
```
bindPoliciesToServiceAccount -> ensurePolicyBindingsFromServiceAccount -> bindPolicyToServiceAccount

unbindPoliciesFromServiceAccount -> removePolicyBindingsFromServiceAccount -> unbindPolicyFromServiceAccount
```
I believe this would be more consistent and easier to follow.

PR #888, opened 2025-10-02
OCM-18460 | wif pool usage after update

miguelhbrito:
mipereir@mipereir-thinkpadp1gen4i:~/myprojects/ocm-cli$ ./ocm gcp update wif-config 2llip7ahv666hld14amh1c7832mcss6j --federated-project sda-ccs-3
....
2025/10/02 14:59:07 Verifying wif-config '2llip7ahv666hld14amh1c7832mcss6j'...
2025/10/02 14:59:17 Ensuring new workload identity pool has recent traffic...
2025/10/02 14:59:18 Trying again in 1 seconds...
2025/10/02 14:59:20 Trying again in 2 seconds...
2025/10/02 14:59:22 Trying again in 4 seconds...
2025/10/02 14:59:27 Trying again in 8 seconds...
2025/10/02 14:59:36 Trying again in 16 seconds...
2025/10/02 14:59:52 Trying again in 32 seconds...
2025/10/02 15:00:25 Checking if the old workload identity pool is still there and not being used...
2025/10/02 15:00:26 wif-config '2llip7ahv666hld14amh1c7832mcss6j' updated successfully.


pkg/gcp/client.go
```diff
@@ -250,3 +259,60 @@ func (c *gcpClient) UpdateWorkloadIdentityPoolOidcIdentityProvider(
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get credentials")
+	}
+
+	// Create authenticated HTTP client
+	client := &http.Client{
+		Transport: &oauth2.Transport{
+			Source: creds.TokenSource,
+		},
+	}
```
rcampos2029, line 281: Having the highlighted code in this method means that a new client is made everytime this method is called. Instead, move this logic to the initializer of the gcpClient object. Call it monitoringClient, and then use this member of the object to perform the query.

cmd/ocm/gcp/update-wif-config.go
```diff
@@ -197,6 +197,42 @@ func updateWorkloadIdentityConfigurationCmd(cmd *cobra.Command, argv []string) e
 			"or contact Red Hat support.", wifConfig.ID())
 	}
 
+	// If the WIF pool is hosted in a different project than the main project, we need to verify the pool usage
+	if wifConfig.Gcp().FederatedProjectId() != wifConfig.Gcp().ProjectId() {
```
rcampos2029, line 201: Let's only run this check if the wif-config was updated on this update call. Also, let's add a flag to skip this check.
This initial check should therefore be:
`if wifProjectUpdated() && !skipProjectVerfication()`

cmd/ocm/gcp/update-wif-config.go
```diff
@@ -197,6 +197,42 @@ func updateWorkloadIdentityConfigurationCmd(cmd *cobra.Command, argv []string) e
...
+		// Check if there are associated clusters first
+		clustersAssociated, err := gcpClientWifConfigShim.HasAssociatedClusters(ctx, connection, wifConfig.ID())
+		if err != nil {
+			return errors.Wrapf(err, "failed to check for associated clusters")
```
rcampos2029, line 205:
```suggestion
			return errors.Wrapf(err, "failed to check if wif-config '%s' has associated clusters", wif.ID())
```

rcampos2029: For user clarity.

cmd/ocm/gcp/update-wif-config.go
```diff
@@ -197,6 +197,42 @@ func updateWorkloadIdentityConfigurationCmd(cmd *cobra.Command, argv []string) e
...
+			log.Println("Ensuring new workload identity pool has recent traffic...")
+			if err := utils.RetryWithBackoffandTimeout(func() (bool, error) {
+				if validationErr := gcpClientWifConfigShim.ValidateNewWifConfigPoolUsage(ctx, wifConfig); validationErr != nil {
+					return true, validationErr
+				}
+				return false, nil
+			}, IamApiRetrySeconds, log); err != nil {
+				return fmt.Errorf("Timed out verifying workload identity pool usage\n"+
+					"Please try 'ocm gcp update wif-config %s' again in a few minutes, "+
+					"or contact Red Hat support.", wifConfig.ID())
```
rcampos2029, line 219: Let's not error out here. The update may still have been successful, we just cannot confirm that the wif usage has moved from the old project to the new one. We should therefore just log a warning.

rcampos2029:
```suggestion
				log.Warning("Timed out verifying workload identity pool usage\n"+
				   "Cannot determine if old identity pool may be removed." +
				   "Please consult user documentation for manually checking usage")
				return nil
```

cmd/ocm/gcp/update-wif-config.go
```diff
@@ -197,6 +197,42 @@ func updateWorkloadIdentityConfigurationCmd(cmd *cobra.Command, argv []string) e
...
+			if err := utils.RetryWithBackoffandTimeout(func() (bool, error) {
+				if err := gcpClientWifConfigShim.ValidateOldWifConfigPoolUsage(ctx, wifConfig); err != nil {
+					return true, err
+				}
+				return false, nil
+			}, IamApiRetrySeconds, log); err != nil {
+				return fmt.Errorf("Timed out verifying old workload identity pool usage\n"+
+					"Please try 'ocm gcp update wif-config %s' again in a few minutes, "+
+					"or contact Red Hat support.", wifConfig.ID())
+			}
```
rcampos2029, line 232: Same comment as when checking new pool.

cmd/ocm/gcp/update-wif-config.go
rcampos2029: Also, please move the verification logic into a separate helper. This is to reduce the size of this method and for potential reuse of this functionality.

rcampos2029 (changes requested):

cmd/ocm/gcp/update-wif-config.go
rcampos2029: Also, this error should be actionable. In this case, the user action would be to try again or contact support.

cmd/ocm/gcp/update-wif-config.go
rcampos2029: `--skip-cross-project-verification` - Bypasses check when updating the federated project

PR #1078, opened 2026-04-03
OCM-23390 | feat: prefetch machine types and versions in interactive mode

rcampos2029:
## Description
Implement background prefetching for machine types and version options in interactive cluster creation to reduce user wait time.

## Problem
In interactive mode for `ocm create cluster`, users experience delays when prompted for machine types and versions because these API calls can take several seconds to complete.

## Solution
Prefetch these slow API calls in the background:
- Machine types are prefetched as soon as provider and CCS are determined
- Versions are prefetched as soon as authentication is configured
- Results are cached and ready when the user reaches those prompts

## Benefits
- Significantly reduced wait time in interactive mode
- Better user experience with smoother flow
- No changes to non-interactive mode behavior

## Technical Details
- Added prefetchCache structure with mutex-protected goroutines
- Prefetch starts early in the flow but results are consumed later
- Falls back to synchronous fetch if prefetch hasn't completed (edge case)

## Changes
Modified `cmd/ocm/create/cluster/cmd.go`:
- Added prefetchCache structure for thread-safe caching
- Created background fetch functions for versions and machine types
- Integrated prefetching into preRun() workflow
- Updated consumption points to use cached results

## Testing
- Tested interactive cluster creation flow
- Verified prefetch improves response time
- Confirmed fallback to synchronous fetch works correctly

## Related Jira
https://issues.redhat.com/browse/OCM-23390

### DIFF
--- a/cmd/ocm/create/cluster/cmd.go
+++ b/cmd/ocm/create/cluster/cmd.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cluster
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -25,6 +26,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/AlecAivazis/survey/v2"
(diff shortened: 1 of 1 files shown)


cmd/ocm/create/cluster/cmd.go
```diff
@@ -731,12 +900,22 @@ func minComputeNodes(ccs bool, multiAZ bool) (min int) {
...
 	// Create the client for the OCM API:
 	connection, err := ocm.NewConnection().Build()
 	if err != nil {
 		return fmt.Errorf("Failed to create OCM connection: %v", err)
 	}
-	defer connection.Close()
+	defer func() {
+		// Cancel any background prefetch operations and close connection
+		// Don't wait for goroutines to complete - early exits should not block on speculative work
+		prefetch.cancel()
```
miguelhbrito, line 916:
```suggestion
		// Cancel any background prefetch operations and wait for goroutines to finish
		// before closing the connection they may be using
		prefetch.cancel()
		prefetch.wg.Wait()
```

miguelhbrito (changes requested):

miguelhbrito (approved):

